PR #73, opened 2025-08-18
Optimisation of the configuration of AutomatonChip

irakoton:
Optimisation of the base algorithms managing automata to reduce the configuration time of the automaton chip.

1. A significant performance bottleneck was a single line of the minimisation algorithm (the main loop of `automaton::nerode_congruence` was recursing a lot through a vector with a lot of duplicates). When removing duplicates with a HashSet conversion, configuration time went from 28 s to 6 s for the JWT benchmark.

2. Reworking the base data structure `RawAutomaton`. Now:
a.  states are always encoded as usize to have light memory representations
b. epsilon transitions are removed from the model (there closure is directly added when an epsilon transition was added before)
c. Some basic constructions (language iteration, union, concatenation) distinguish more frequent patterns that lead to a smaller translation Regex -> Automaton. This has a significant impact on the cost of minimisation (for JWT for example, the minimisation was compressing 18k transitions into 5k, now it is 8k to 5k.
d. HashMap/Set are replaced by FxHashMap and FxHashSet due to the absence of adversarial inputs.
e. How transitions are stored is different. Instead of storing raw tuples `(source state, letter, target state)`, they are collected into an organised vector `v : Vec<Vec<(letter, target state)>>`, where `v[s]` is the vector of `(letter,target state)` pairs that have `s` as a source state. In many cases, this replaces a full iteration over transitions by a single index access.

All optimisations of 2. decreased the configuration time for the JWT benchmark from 6 s to 0.4 s.

This PR also introduces serialisation of the entire parsing library + consistency checks based on recomputing all automaton parsers. For the JWT benchmark, the serialised automaton is 131 KB and is included in the code files using a `include_bytes!`.

With the above optimisations, it seems that regarding configuration of the automaton chip, one parser instance is a trade-off between ~1 s of computation or ~200 KB of embedded serialised data.


circuits/src/parsing/automaton.rs
```diff
@@ -99,177 +99,120 @@ impl Letter {
...
     /// out any transition involving a byte outside of the alphabet.
-    pub(super) fn singleton(letters: &[u8], alphabet_size: usize) -> RawAutomaton<usize> {
+    pub(super) fn byte_concat(word: &[Vec<Letter>], alphabet_size: usize) -> Self {
+        let mut transitions = Vec::from_iter(
+            word.iter()
+                .map(Vec::len)
+                .map(Vec::with_capacity)
+                .chain(once(vec![])),
+        );
+        let mut markers = FxHashSet::with_capacity_and_hasher(2, FxBuildHasher);
```
iquerejeta, line 184: Why are you initialising the HashSet with capacity only 2?

circuits/src/parsing/automaton.rs
```diff
@@ -213,64 +217,128 @@ impl RawAutomaton {
...
-    fn live_states(&self) -> HashSet<State> {
-        let reach_states = self.reachable_states();
-        let mut back_reach_states = HashSet::new();
-        let mut visited = HashSet::new();
-        let mut pending_states = Vec::with_capacity(self.nb_states);
+impl RawAutomaton {
+    /// Computes the simplified reachability graph of an automaton, i.e., the
+    /// transitions without letters. Computing once and for all a simplified
+    /// graph tends to make the code more efficient when the transition table
+    /// has be traversed many times, since there are often 10~100 transitions
```
iquerejeta, line 281:
```suggestion
    /// has to be traversed many times, since there are often 10~100 transitions
```

circuits/src/parsing/serialization.rs
```diff
@@ -0,0 +1,351 @@
...
+// serialization of the `RawAutomaton` type, this would only save two bytes per
+// serialized automaton, which is quite insignificant.
+impl Serialize for bool {
+    #[cfg(test)]
+    fn serialize(&self, buf: &mut Vec<u8>) {
+        if *self { 1u8 } else { 0u8 }.serialize(buf)
+    }
+    fn deserialize(buf: &mut &[u8]) -> Result<Self, String> {
+        Ok(u8::deserialize(buf)? == 1)
+    }
```
iquerejeta, line 109: any value other than 1 will return false. Is that expected? If yes, you could also avoid implementing Serialize for `bool`, and treating them as `u8`s.

circuits/src/parsing/mod.rs
```diff
@@ -27,6 +27,8 @@ pub mod automaton_chip;
 /// A module to specify languages as regular expressions and convert them into
 /// finite automata.
 pub mod regex;
+/// A module to serialize automata as sequences of bytes.
+pub mod serialization;
```
iquerejeta: I would make this private to the automaton.

iquerejeta (approved):
LGTM - great optimisations 👏

circuits/src/parsing/regex.rs
```diff
@@ -121,9 +97,9 @@ where
     /// belong to a given collection. Is equivalent to
     /// `Self::any_byte().minus(Self::byte_from(l))` but is more efficient.
     fn byte_not_from(l: impl IntoIterator<Item = u8>) -> Self {
-        let mut bytes: [bool; ALPHABET_MAX_SIZE - 1] = core::array::from_fn(|_| true);
+        let mut bytes: [bool; ALPHABET_MAX_SIZE] = core::array::from_fn(|_| true);
         l.into_iter().for_each(|b| bytes[b as usize] = false);
-        Self::byte_from((0..(ALPHABET_MAX_SIZE - 1) as u8).filter(|&b| bytes[b as usize]))
+        Self::byte_from((0..=(ALPHABET_MAX_SIZE - 1) as u8).filter(|&b| bytes[b as usize]))
```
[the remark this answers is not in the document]
irakoton: I think the current version is necessary, since ALPHABET_MAX_SIZE = 256. If I'm correct, casting it as u8 overrides it to 0.

circuits/src/parsing/automaton.rs
```diff
@@ -300,40 +368,52 @@ impl RawAutomaton {
...
+        let (transitions, markers) = RawAutomaton::filter_map_transitions(
+            &self.transitions,
+            |state| renaming.get(&state).copied(),
+            live_states.len(),
+            0,
+        );
+
+        Self {
+            deterministic: self.deterministic,
+            complete: false, // `false` most of the time.
```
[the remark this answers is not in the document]
irakoton: No, but it was indeed confusing. I made a more complete comment in the latest force push to explain what I meant.

PR #90, opened 2025-08-29
Load only necessary bit lengths in pow2range chip

miguel-ambrona:
We revisit the `FromScratch` trait, concretely, the signature of `load_from_scratch`.

We simplify the `pow2range` chip in such a way that only necessary bit lengths are loaded. For this, we keep track of the enforced bit lengths before loading the table. This requires loading the table at the end of a synthesize instead of at the beginning. We now do that for all chips.


iquerejeta (approved):
Very clean - LGTM 👍

circuits/examples/ivc.rs
```diff
@@ -50,6 +50,12 @@ type CBase = <C as CircuitCurve>::Base;
 
 type NG = NativeGadget<F, P2RDecompositionChip<F>, NativeChip<F>>;
 
+#[cfg(feature = "truncated-challenges")]
+const K: u32 = 18;
+
+#[cfg(not(feature = "truncated-challenges"))]
+const K: u32 = 19;
+
```
irakoton, line 58: Is there a reason why these values are increased from 16 to 18-19? I don't get which part of the changes required this, nor how we still get better golden files despite this.

circuits/src/biguint/biguint_gadget.rs
```diff
@@ -794,11 +794,8 @@ where
         <N as FromScratch<F>>::configure_from_scratch(meta, instance_columns)
     }
 
-    fn load_from_scratch(
-        layouter: &mut impl Layouter<F>,
-        config: &Self::Config,
-    ) -> Result<(), Error> {
-        <N as FromScratch<F>>::load_from_scratch(layouter, config)
+    fn load_from_scratch(&self, layouter: &mut impl Layouter<F>) -> Result<(), Error> {
+        self.native_gadget.load_from_scratch(layouter)
```
irakoton: This is the change I like the most honestly. Taking the chip as argument feels much more natural.

circuits/goldenfiles/examples/PoseidonExample
```diff
@@ -1,7 +1,7 @@
 CircuitModel {
-    k: 10,
+    k: 6,
     rows: 46,
-    table_rows: 511,
+    table_rows: 1,
```
irakoton, line 4: Now I'm starting to get confused, how is such an extreme improvement possible? From the numbers, it seems like a `load` isn't called anymore in this case, but I fail to find the modification causing this.

irakoton (approved):
LGTM, no specific requests on my side. The interface feels much more natural this way. I would like to better understand what triggers the improvements in some of the goldenfile examples though.

circuits/goldenfiles/examples/PoseidonExample
miguel-ambrona: The new table only includes the range-check tags that are used.
I believe that test is not using the pow2range table at all, thus the table is filled with 1 row only (the default case for disabling the lookup).

PR #100, opened 2025-09-09
fix: clippy

gilescope:
clippy in ci was silently failing claiming it wasn't installed and needed adding.


iquerejeta (approved):
LGTM - why are changelogs now required?

gilescope:
Don't merge yet. Will comment when it's ready.

PR #102, opened 2025-09-15
Expose function for determining number of deserialised points and size of MSM

iquerejeta:
The number of points involved in a proof allows one to compute ~80% of the cost of the verifier cost. This function exposes the number of points in the verifying key and in the proof.


miguel-ambrona (approved):
LGTM!

PR #104, opened 2025-09-17
Design a layouter that computes `k` and the number of instances without hints

iquerejeta:
The cost model now computes the value of `k` for a given circuit without looping over the synthesize function.


circuits/src/compact_std_lib.rs
```diff
@@ -1360,7 +1360,7 @@ impl<Rel: Relation> MidnightPK<Rel> {
 ///     }
 /// }
 ///
-/// const K: u32 = 13;
+/// const K: u32 = 14;
```
iquerejeta, line 1363: Why is this? - review

miguel-ambrona: It's probably related to this:
https://github.com/midnightntwrk/midnight-zk/pull/103/commits/c39901c3fc92fbcb855475a8b6301d2737949f5d

PR #106, opened 2025-09-17
Compute optimal `max_bit_len` in `compact_std_lib`

miguel-ambrona:
Now that the cost model is so cheap, we can afford trying different choices of `max_bit_len` and keep the one that minimizes `k`.

### DIFF
--- a/circuits/src/compact_std_lib.rs
+++ b/circuits/src/compact_std_lib.rs
@@ -1148,10 +1148,9 @@ impl<'a, R: Relation> MidnightCircuit<'a, R> {
 
         let mut best_k = u32::MAX;
         let mut best_max_bit_len = 8;
-        let mut max_bit_len = max_bit_len_opt.unwrap_or(8);
 
         // Loop for finding the optimal `max_bit_len`.
-        loop {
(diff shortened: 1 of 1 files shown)


circuits/src/compact_std_lib.rs
```diff
@@ -1135,21 +1125,63 @@ pub struct MidnightCircuit<'a, R: Relation> {
...
+                max_bit_len,
+                instance: Value::unknown(),
+                witness: Value::unknown(),
+                nb_public_inputs: Rc::new(RefCell::new(None)),
+            })
+        };
+
+        let mut best_k = u32::MAX;
+        let mut best_max_bit_len = 8;
+        let mut max_bit_len = max_bit_len_opt.unwrap_or(8);
```
iquerejeta, line 1161: You only reach this code if max_bit_len_opt is None, so you can define it directly as 8.

iquerejeta (approved):

circuits/src/compact_std_lib.rs
miguel-ambrona: Addressed!

PR #109, opened 2025-09-19
Cache range-checked cells

miguel-ambrona:
We cache range-checked cells so that they do not need to be range-checked again when it can be avoided.


circuits/src/field/native/native_gadget.rs
```diff
@@ -261,6 +272,13 @@ where
         x: &AssignedNative<F>,
         bound: &BigUint,
     ) -> Result<(), Error> {
+        if let Some(current_bound) = self.constrained_cells.borrow().get(x) {
+            if current_bound <= bound {
+                return Ok(());
+            }
+        }
```
iquerejeta, line 279: The bound in `AssignedBounded` is non inclusive, i.e. `[0, 2^bound)`. You should use `if current_bound < bound`

circuits/src/field/native/native_gadget.rs
```diff
@@ -771,6 +795,7 @@ where
         _layouter: &mut impl Layouter<F>,
         byte: &AssignedByte<F>,
     ) -> Result<AssignedNative<F>, Error> {
+        self.update_bound(&byte.0, BigUint::from(256u32));
```
iquerejeta, line 798: Why don't you use the cache when you are assigning types such as `Byte` and `Bit`?

circuits/src/field/native/native_gadget.rs
miguel-ambrona: I think it is correct the way it is.

`assert_lower_than_fixed` is also strict.

iquerejeta: Correct 👍

iquerejeta (approved):
LGTM 👍

PR #116, opened 2025-10-06
Run serialization tests single threaded

iquerejeta:
Running tests in a single thread seems to fix the issue. Let's see if this is the final PR wrt this issue.


miguel-ambrona (approved):
LGTM

PR #126, opened 2025-10-20
ZKIR new operations

miguel-ambrona:
- Add
- AssertEqual
- Loading constants

I intentionally left something that I got wrong on my first attempt, to motivate reviewers. Good luck finding it!

### DIFF
--- a/zkir/src/utils/constants.rs
+++ b/zkir/src/utils/constants.rs
@@ -18,7 +18,7 @@
 //!  - A string prefixed by "Jubjub:" is parsed as an [IrType::JubjubPoint]. The
 //!    payload that follows this prefix is expected to be a hexadecimal string of
 //!    exactly 32 bytes, following the repr_J encoding as defined in [Zcash specification
-//!    5.4.9.3 - JubJub](https://zips.z.cash/protocol/protocol.pdf#jubjub).//!
+//!    5.4.9.3 - JubJub](https://zips.z.cash/protocol/protocol.pdf#jubjub).
 //!    Additionally, the following strings are allowed:
 //!      - "Jubjub:GENERATOR", for the official generator of the Jubjub subgroup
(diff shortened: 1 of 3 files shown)


zkir/src/utils/constants.rs
```diff
@@ -0,0 +1,220 @@
...
+        "GENERATOR" => Ok(JubjubSubgroup::generator()),
+        "IDENTITY" => Ok(JubjubSubgroup::identity()),
+        _ => {
+            let bytes: [u8; 32] = const_hex::decode(str)
+                .map_err(|e| Error::Other(format!("{e:?}")))?
+                .try_into()
+                .map_err(|_| Error::ParsingError(IrType::JubjubPoint, str.to_string()))?;
+            JubjubExtended::from_bytes(&bytes)
+                .into_option()
+                .map(|p| p.into_subgroup())
```
iquerejeta: Do we want to allow points outside the subgroup? I think its safer to return an error. This is probably not intended.

miguel-ambrona: I'm not sure what you mean.
This is the "official" way to build a JubjubSubgroup from bytes.

zkir/src/utils/constants.rs
```diff
@@ -0,0 +1,220 @@
+//! This module defines how constant values must be parsed.
+//!
+//!  - A single character is parsed as an [IrType::Bool]: '0' is false, '1' is
+//!    true and any other character results in an error.
+//!
+//!  - A hexadecimal string is parsed as [IrType::Bytes].
```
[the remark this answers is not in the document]
iquerejeta: Not in spanish.

zkir/src/utils/constants.rs
iquerejeta: Ah, I see that `into_subgroup` fails if the point is not in the subgroup. That's what I meant. All good 👍

PR #128, opened 2025-10-21
ZKIR yet more operations

miguel-ambrona:
- IsEqual
- Mul
- Sub
- Neg
- InnerProduct

### DIFF
--- a/zkir/src/types.rs
+++ b/zkir/src/types.rs
@@ -127,7 +127,7 @@ macro_rules! impl_enum_from_try_from {
                 fn try_from(value: $enum) -> Result<Self, Self::Error> {
                     match &value {
                         $enum::$variant(inner) => Ok(inner.clone()),
-                        other => Err(Error::ExpectingType(value.get_type(), other.get_type())),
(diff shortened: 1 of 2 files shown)


zkir/src/instructions/operations/is_equal.rs
```diff
@@ -0,0 +1,51 @@
...
+    use CircuitValue::*;
+    let b = match (x, y) {
+        (Bool(a), Bool(b)) => std_lib.is_equal(layouter, a, b)?,
+
+        (Bytes(v), Bytes(w)) if v.len() == w.len() => {
+            let pair_wise_eq = (v.iter().zip(w))
+                .map(|(vi, wi)| std_lib.is_equal(layouter, vi, wi))
+                .collect::<Result<Vec<_>, plonk::Error>>()?;
+            std_lib.and(layouter, &pair_wise_eq)?
+        }
```
iquerejeta: This can be confusing. If the size of the two vectors are not the same, you get an error `Err(Error::Unsupported(Operation::IsEqual, x.get_type()))`. The error should be different here.

PR #137, opened 2025-10-31
Picus extraction support

0xddom:
This PR includes the changes made to the original crates in order to support the extraction as well as additional crates that mainly define the interfaces expected by the extractor.

Changes can be categorized in two main groups:
- Support for defining groups inside the circuit
- Support for integrating with the extractor

The former comprises the changes made to the `proofs` crate as well as any application of the `#[picus::group]` or `#[derive(picus::DecomposeInCells)]` macros.

The latter comprises the `extractor/support` crate, and any changes in the `circuits` crate encapsulated in a module (usually named `extraction`) guarded by the `extraction` feature flag.

I added comments in parts of the code explaining things like the rationale behind certain changes or architecture explanations that help understand it.


circuits/src/field/decomposition/pow2range.rs
```diff
@@ -237,6 +237,52 @@ impl<F: PrimeField> Pow2RangeChip<F> {
...
+        }
+
+        fn load_chip(&self, layouter: &mut L, _config: &Self::Config) -> Result<(), Error> {
+            self.load_table(layouter)
+        }
+    }
+
+    impl<F: PrimeField> ChipArgs for Pow2RangeChip<F> {
+        type Args = usize;
+    }
```
0xddom: For enabling extraction for a chip these two traits need to be implemented. The general structure mimics `FromScratch` and `ComposableChip`. Consequently, most implementations of these traits can be derived from an existing implementation of `FromScratch` and I included two derive macros that automatically implements these traits if the chip implements `FromScratch`.

circuits/src/ecc/hash_to_curve/htc_gadget.rs
```diff
@@ -32,6 +32,12 @@ use crate::{
 };
 
 #[derive(Clone, Debug)]
+#[cfg_attr(
+    feature = "extraction",
+    derive(picus::NoChipArgs, picus::InitFromScratch),
+    from_scratch(H),
+    from_scratch(E)
+)]
```
0xddom: One instance of the derive macros. `InitFromScratch` is configurable with a couple helper attributes (docs about the macro can be found [here](https://docs.rs/mdnt-support-macros/latest/mdnt_support_macros/#derive-macros)).

proofs/src/circuit/groups.rs
```diff
@@ -0,0 +1,1143 @@
+//! Traits and structs related to groups of regions.
```
0xddom: This module has all the support types and traits for the group API. More than half of it are unit tests so is less code than it seems at first. 

Also, this file alone accounts for a third of the PR

proofs/src/plonk/circuit.rs
```diff
@@ -97,6 +97,18 @@ impl<C: ColumnType> PartialOrd for Column<C> {
...
+#[cfg(feature = "extraction")]
+pub mod extraction {
+    //! Extraction support for column types.
+
+    use extractor_support::{auto_conf_impl, circuit::configuration::AutoConfigure};
+
+    auto_conf_impl!(super::Column<super::Fixed>, fixed_column, crate);
+    auto_conf_impl!(super::Column<super::Instance>, instance_column, crate);
+    auto_conf_impl!(super::Column<super::Advice>, advice_column, crate);
+    auto_conf_impl!(super::TableColumn, lookup_table_column, crate);
```
0xddom: This trait is used by the extractor while working with the chip's configuration. Allows for passing a `&mut ConstraintSystem<F>` between complex types and needs leaf implementations for these types.

- [Docs for the trait](https://docs.rs/mdnt-support/latest/mdnt_support/circuit/configuration/trait.AutoConfigure.html)
- [Docs for the macro](https://docs.rs/mdnt-support/latest/src/mdnt_support/circuit/configuration.rs.html#14-25)

PR #138, opened 2025-11-10
Implement parallel computation for running product polynomial

John-Gong-Math:
Although the improvement regarding computing the running product is significient (`2 ~ 4` times speedup when `k = 11, .. 16`), the `commit()` function of permutation argument almost remains the same, since the cost of running product computation is not the bottleneck. Therefore, we discard this PR. See below the benchmark about `zswap circuit`:

- with parallel computation of running product:
   ZSwap Prover/Commit permutation functions
                        time:   [77.289 ms 77.443 ms 77.613 ms]

- initial implementation: 
   ZSwap Prover/Commit permutation functions
                        time:   [78.203 ms 78.444 ms 78.722 ms]

### DIFF
--- a/proofs/Cargo.toml
+++ b/proofs/Cargo.toml
@@ -32,6 +32,10 @@ name = "zswap_output"
 required-features = ["bench-internal"]
 harness = false
 
+[[bench]]
+name = "par_running_prod"
+harness = false
+
 [dependencies]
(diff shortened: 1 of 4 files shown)


John-Gong-Math:
Screenshot 2025-11-13 at 6 15 22 PM

PR #146, opened 2025-11-24
Iquerejeta/build fixes

### LINKED ISSUE #129: Cargo test --all-features error
### Context & versions:

Latest `main` (as of a01ff713a69c0ae29c929ab41be262337d5670d3)

### Steps to reproduce & Behaviour

Running `cargo test --all-features` in the `circuits` and `proofs` dir produces several errors.

iquerejeta:
Closes #129 

Removing some unused features, and fixing the different crates to allow a `cargo build --all-features` from the root folder. 

I've changed the aggregator so that it does not compile if the `truncated-challenges` feature is active (which is not supported) and prints a warning message.


irakoton (approved):
LGTM. Good that you've changed the `format!` to `to_string`, that's a warning I've been adding to previous PRs for a while, but they never got merged.

PR #161, opened 2025-12-10
Blind limb of quotient polynomials and ensure ZK

### LINKED ISSUE #97: Blind limbs of quotient polynomial (for ensuring ZK property)
Problem statement
--------------------------------
The prover divides the quotient poly into limbs, and commits to each limb separately. For being able to simulate proofs, these limbs need to be blinded.

Suggested resolution
--------------------------------
Blind each limb, while making sure the effect of each blinding cancels out when re-constructing the quotient poly from its limbs.

remalue:
Fixes #97

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,6 +123,16 @@ jobs:
           package: midnight-aggregator
           args: --release --all-features
 
+  test-zkir:
+    if: github.event.pull_request.draft == false
+    name: Test ZKIR
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3
+      - uses: ./.github/actions/cargo-test
+        with:
+          package: midnight-zkir
+          args: --release --all-features
 
   doc-links:
     if: github.event.pull_request.draft == false
--- a/zkir/tests/operations.rs
+++ b/zkir/tests/operations.rs
@@ -96,7 +96,7 @@ fn test_publish() {
     // Variable resolution: Published variables must be defined.
     test_without_witness(
         &[(Publish, vec!["x"], vec![])],
-        Some(Error::ParsingError(IrType::Bool, "x".to_string())),
+        Some(Error::NotFound("x".to_string())),
     );
     test_without_witness(
         &[


circuits/src/utils/plonk_api.rs
```diff
@@ -261,14 +261,19 @@ pub fn update_circuit_goldenfiles<R: Relation>(relation: &R) {
 
 /// Use filecoin's SRS (over BLS12-381)
 pub fn filecoin_srs(k: u32) -> ParamsKZG<Bls12> {
-    assert!(k <= 19, "We don't have an SRS for circuits of size {k}");
+    assert!(
+        k <= 18,
```
iquerejeta: We don't have for k = 19?

miguel-ambrona:
> Why do you need to increase the degree of the polynomial? Why not blind them as follows (as described in the paper):
> 
> ```
> h(X) = t'_0(X) + X^N * t'_1(X) + X^2N * t'_2(X)
> 
> t_0(X) = t'_0(X) + b_0 X^N
> t_1(X) = t'_1(X) - b_0 + b_1 X^N
> t_2(X) = t'_2(X) - b_1
> 
> with b_0, b_1 random
> ```

@iquerejeta Note that this approach requires committing to a polynomial of degree `N` (due to the term `X^N`).
Before this, we were committing to polys of at most `N-1`.

In order not to increase the SRS degree, we can:
 - Split in more pieces (1 more piece)
 - Use the fact that the quotient polynomial does not have full degree, there is a small room for some blinding factors between it and the next power of 2.
 
 I suggest going for this second approach.

proofs/src/plonk/vanishing/prover.rs
```diff
@@ -128,6 +130,50 @@ impl<F: WithSmallOrderMulGroup<3>> Committed<F> {
...
 
+fn blind_even_part<F: PrimeField>(blinding_scalar: F, limbs: &mut [Polynomial<F, Coeff>]) {
+    for (index, piece) in limbs.iter_mut().enumerate() {
+        match index % 2 {
+            0 => piece.values.push(blinding_scalar),
+            1 => piece.values[0] -= blinding_scalar,
+            _ => {}
+        }
+    }
+}
```
miguel-ambrona: Why do you use the same blinding scalar for all pieces? They should be independent (but chained).

proofs/src/plonk/vanishing/prover.rs
```diff
@@ -128,6 +130,50 @@ impl<F: WithSmallOrderMulGroup<3>> Committed<F> {
...
+                2 => piece.values[0] -= rho1,
+                _ => {}
+            }
+        }
+        quotient_limbs.extend(even_part);
+
+        quotient_limbs
+    }
+}
+
```
miguel-ambrona:
```suggestion
// This function requires the quotient limbs to be in little-endian order.
fn blind_quotient_limbs<F: PrimeField>(quotient_limbs: &mut [Polynomial<F, Coeff>]) {
    assert!(quotient_limbs.len() >= 2);

    for i in 1..quotient_limbs.len() {
        let t = F::random(OsRng);
        quotient_limbs[i - 1].values.push(t);
        quotient_limbs[i].values[0] -= t;
    }
}

```

proofs/src/plonk/vanishing/prover.rs
remalue: Exactly!

proofs/src/plonk/vanishing/prover.rs
remalue: I am aware that my initial approach to blinding is not enough to devise a simulator and have already worked on a proper solution.

proofs/src/plonk/vanishing/prover.rs
```diff
@@ -128,6 +130,50 @@ impl<F: WithSmallOrderMulGroup<3>> Committed<F> {
...
+            0 => piece.values.push(blinding_scalar),
+            1 => piece.values[0] -= blinding_scalar,
+            _ => {}
+        }
+    }
+}
+
+fn blind_quotient_limbs<F: PrimeField>(
+    mut quotient_limbs: Vec<Polynomial<F, Coeff>>,
+) -> Vec<Polynomial<F, Coeff>> {
```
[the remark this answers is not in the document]
remalue: Yes, I am aware. The current approach is just meant as a discussion starter. Let's focus on the proper blinding first, then I polish.

proofs/src/plonk/vanishing/prover.rs
```diff
@@ -131,6 +133,19 @@ impl<F: WithSmallOrderMulGroup<3>> Committed<F> {
...
+    let nr_limbs = quotient_limbs.len();
+    assert!(nr_limbs >= 2);
+
+    for i in 1..nr_limbs {
+        let t = F::random(OsRng);
+        quotient_limbs[i - 1].push(t);
+        quotient_limbs[i][0] -= t;
+    }
+
+    quotient_limbs[nr_limbs - 1].push(F::ZERO);
```
miguel-ambrona: Is this necessary?

proofs/src/plonk/vanishing/prover.rs
```diff
@@ -103,15 +103,18 @@ impl<F: WithSmallOrderMulGroup<3>> Committed<F> {
...
-            .map(|v| domain.coeff_from_vec(v.to_vec()))
+        let mut h_pieces = h_poly
+            .chunks_exact((domain.n - 1) as usize)
+            .map(|v| v.to_vec())
             .collect::<Vec<_>>();
         drop(h_poly);
 
+        let h_pieces: Vec<_> =
+            h_pieces.into_iter().map(|h_piece| domain.coeff_from_vec(h_piece)).collect();
+
```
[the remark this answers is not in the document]
remalue: Creating it directly doesn't work. We'd need to pad the chunks first. `coeff_from_vec` has this check: `assert_eq!(values.len(), self.n as usize)`, which fails without padding.

I think it's nicer to pad only the most significant limb instead of padding all limbs before blinding and then modifying their inner values.

proofs/src/plonk/vanishing/prover.rs
remalue: See above.

PR #163, opened 2025-12-12
Exposing variable-length poseidon and sha256

irakoton:
Simply exposing the variable-length versions of sha256 and Poseidon, which were available in `main` but not in `compact_std_lib`. The configuration is done in a way that activating sha256 / poseidon automatically activates its variable-length counterpart.


irakoton:
Exposing these functions is not scheduled to happen before mainnet.

PR #169, opened 2025-12-18
Release new version of several crates

iquerejeta:
This PR falls back to the relative path imports of workspace dependencies. We, anyways, had to patch the libraries in a previous PR when using external libraries. This also facilitates the release of versions.

This PR releases new versions for `proofs, circuits` and `zk_stdlib`.


miguel-ambrona (approved):
LGTM.

PR #176, opened 2025-12-24
proofs: expose the "powers of s" of KZG params

miguel-ambrona:


proofs/src/poly/kzg/params.rs
```diff
@@ -117,6 +117,13 @@ impl<E: Engine + Debug> ParamsKZG<E> {
         }
     }
 
+    /// Returns the "powers of s" of these KZG params, that is,
```
iquerejeta:
```suggestion
    /// Returns the committed "powers of s" of these KZG params, that is,
```

PR #181, opened 2026-01-19
Curve25519

davidnevadoc:
Add Curve25519  wrapping curve25519_dalek. 
  ## Changes

###  Curves
  - Wrapper around `EdwardsPoint` (`Curve25519`) and subgroup type  (`Curve25519Subgroup`). 
 Needed because `group::Curve` cannot be   implemented for the foreign `EdwardsPoint` directly.
  - Wrapper around the affine representation (`Curve25519Affine`).
  Needed because curve25519_dalek does not expose point coordinates publicly; we recover them from the compressed encoding and cache the result.
  - Base field (`Fp`). Expansion of the impl_field macro with a manual square root implementation.
  Necessary because curve25519_dalek does not expose its internal `FieldElement`. May be removed after https://github.com/dalek-cryptography/curve25519-dalek/pull/816

 ### Circuits
  - `CircuitCurve` and `EdwardsCurve` implementations for `Curve25519`.
  - `FieldEmulationParams` for both the base field and scalar field over BLS12-381's scalar field.

 ## Notes

  - batch_normalize uses a naive per-point loop. An optimized version using Montgomery's trick (batch inversion of Zs) is not feasible without access to curve25519_dalek  internals (`FieldElement` and `EdwardsPoint` coordinates are `pub(crate)`).
  - The `Curve25519Affine` struct caches the underlying `EdwardsPoint` to avoid repeated decompression. This can be simplified once  https://github.com/dalek-cryptography/curve25519-dalek/pull/819 exposes the  affine point API.

### DIFF
--- a/curves/src/curve25519/affine.rs
+++ b/curves/src/curve25519/affine.rs
@@ -15,8 +15,7 @@
 
 use curve25519_dalek::{edwards::CompressedEdwardsY, EdwardsPoint};
 use ff::Field;
-use group::Group;
-use group::GroupEncoding;
+use group::{Group, GroupEncoding};
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
 use super::{


curves/src/curve25519/curve.rs
```diff
@@ -0,0 +1,434 @@
...
+impl_edwards_curve_ops!(Curve25519Subgroup);
+
+impl Group for Curve25519Subgroup {
+    type Scalar = Scalar;
+
+    fn random(mut rng: impl rand_core::RngCore) -> Self {
+        // Generate a random point and clear cofactor to ensure it's in subgroup.
+        let point = EdwardsPoint::random(&mut rng);
+        Curve25519Subgroup(point.mul_by_cofactor())
+    }
```
John-Gong-Math: This is minor, but is it better to assert `EdwardsPoint::random(&mut rng)` torsin free, otherwise take another random point instead of multiplying by the cofactor?

curves/src/curve25519/curve.rs
```diff
@@ -0,0 +1,434 @@
...
+                $type(-self.0)
+            }
+        }
+
+        impl Neg for &$type {
+            type Output = $type;
+            fn neg(self) -> $type {
+                $type(-self.0)
+            }
+        }
```
iquerejeta: Duplicated code?

curves/src/curve25519/curve.rs
```diff
@@ -0,0 +1,434 @@
...
+
+/// EdwardsPoint wrapper for circuit integration.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct Curve25519(pub EdwardsPoint);
+
+/// Curve25519 point guaranteed to be in the prime-order subgroup.
+/// This type can only be constructed via `Curve25519::into_subgroup()` which
+/// verifies that the point is torsion-free.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct Curve25519Subgroup(pub(crate) EdwardsPoint);
```
iquerejeta: If we are creating a safe type here, I'd make the visibility of the inner point private. I see you have a `getter` below, so you should not need to make this public. 

Moreover, you document that 'This type can only be constructed via `Curve25519::into_subgroup()`', however, you create it with `from_edwards` function. Please update the docs.

curves/src/curve25519/fp.rs
```diff
@@ -0,0 +1,607 @@
...
+        const EXP: [u64; 4] = [
+            0xfffffffffffffffd,
+            0xffffffffffffffff,
+            0xffffffffffffffff,
+            0x0fffffffffffffff,
+        ];
+        let a1 = self.pow_vartime(EXP);
+        let a0 = (a1.square() * self).square();
+
+        let valid = a0.ct_eq(&-Self::ONE);
```
iquerejeta: rename to `invalid`?

circuits/src/ecc/curves.rs
```diff
@@ -17,8 +17,11 @@ use ff::PrimeField;
 use group::{Curve, Group};
 #[cfg(feature = "dev-curves")]
 use midnight_curves::bn256;
+#[cfg(feature = "curve25519")]
+use midnight_curves::curve25519::{
+    Curve25519, Curve25519Affine, CURVE_A as CURVE25519_A, CURVE_D as CURVE25519_D,
+};
```
iquerejeta: This will be as 'standard' as secp. Do we really want to put it behind a feature?

iquerejeta (approved):
Looks good. Minor comments.

circuits/src/ecc/curves.rs
davidnevadoc: I wasn't aware of that. I will remove the feature then.

John-Gong-Math (approved):

PR #188, opened 2026-01-29
k256 integration. `FieldEncoding` trait

davidnevadoc:
Superseded by https://github.com/midnightntwrk/midnight-zk/pull/201
 Any changes we make to field traits ( even adding one such as `FieldEncoding`) propagates through almost all the code base. So it is probably better to work on a trait that substitutes `PrimeField` and covers all the issues we have.

 --- 
Introduce the trait `FieldEncoding` to provide a unified interface for transforming prime field elements ,**regardless of interal representation** , from and to bytes, bits and big integers.
After this we will be able to remove cleanly the multitude of helper functions that we currently have across `circuits/` and `curves/`. 
Some of which are currently relying on internal representation assumptions of the fields.

 ## For reviewers
 This PR has 3 commits:
  - 1 Add `FieldEncoding` trait.
  - 2 and 3 . Remove the old `EndianRepr` and substitute it with `FieldEncoding` in the curves generated by the macro `impl_field`.
  Note that these curves are secp256k1 and bn256. The first will be removed in a future PR, and bn256 will be feature-gated and only used for testing. 
  
This PR is not targeting main, it targets an intermediate branch.

---
Before merging:
 - [x] Merge #187 
 - [x] Target branch nev@integration.k256 ( #192)
 
 ---
 Next PR:
 #189  
 
 
 --- Future work
  - This may not be a good name. FieldEncoding suggests it could encode any field, yet it doesn't make much sense to have big int transformation functions for non-prime fields.

### DIFF
--- a/curves/src/field_encoding.rs
+++ b/curves/src/field_encoding.rs
@@ -19,7 +19,7 @@
 //!
 //! # Example
 //!
-//! ```ignore
+//! ```
 //! use midnight_curves::{FieldEncoding, Fq};
 //!
 //! let scalar = Fq::from(42u64);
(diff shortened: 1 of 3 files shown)


curves/src/field_encoding.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Converts the field element to a big-endian byte array.
+    fn to_be_bytes(&self) -> Self::Bytes;
+
+    /// Creates a field element from little-endian bytes.
+    ///
+    /// Returns `None` if:
+    /// - The byte slice has incorrect length.
+    /// - The value is not in the canonical range `[0, modulus)`.
+    fn from_le_bytes(bytes: &[u8]) -> Option<Self>;
```
John-Gong-Math: Same as above, can we directly check the input type by: `&[u8; Self::NUM_BYTES]`?

curves/src/field_encoding.rs
```diff
@@ -0,0 +1,141 @@
...
+    ///
+    /// Note: This does **not** perform modular reduction. If you need reduction,
+    /// compute `n % modulus` before calling this method.
+    fn from_biguint(n: &BigUint) -> Option<Self>;
+
+    /// Converts the field element to little-endian bits.
+    ///
+    /// The returned vector has length `BYTE_SIZE * 8`. The least significant
+    /// bit is at index 0.
+    fn to_le_bits(&self) -> Vec<bool> {
```
John-Gong-Math: As above, the output type: `[bool; Self::NUM_BYTES * 8]`?

curves/src/field_encoding.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn to_le_bytes(&self) -> Self::Bytes;
+
+    /// Converts the field element to a big-endian byte array.
+    fn to_be_bytes(&self) -> Self::Bytes;
+
+    /// Creates a field element from little-endian bytes.
+    ///
+    /// Returns `None` if:
+    /// - The byte slice has incorrect length.
+    /// - The value is not in the canonical range `[0, modulus)`.
```
John-Gong-Math: When it comes to `Fq2` or more general extension fields, `modulus` is meant for the size of the whole field, IIRC. I think we'd better specify it as `modulus of the prime subfield`.

PR #191, opened 2026-01-30
k256 integration: base field wrapper

davidnevadoc:
This PR is optional.  We have 2 alternatives for exposing the base field:
 1. Direct re-export from `k256`
 2. Wrapper for the base field (this PR)
 
Directly re-exporting the field makes the user responsible for normalization of the field element. Without normalization the following functionality will either panic (in debug) or return an erroneous result (in release):
  - `is_zero()`, `is_odd()`, `is_even()`, `==`, `!=`, `ct_eq()`
  
 This wrapper simply overwrites these methods and adds a `normalize()` call first to ensure the result is correct.
 
---
Despite looking like a very serious issue, if we used this field directly in `circuits/` only the `is_odd`, `is_even` functions would have been problematic (they are used in map-to-curve impl). The rest of the functions are not used unsafely.

For reviewers
---
This PR has 3 commits:
- 1. Add the wrapper and the methods that need explicit normalization.
- 2. The rest of wrappers that just call the inner type method.
- 3. Tests

Before merging:
---

 - [x] Merge https://github.com/midnightntwrk/midnight-zk/pull/189
 - [x] Target branch nev@integration.k256 ( https://github.com/midnightntwrk/midnight-zk/pull/192)
 
 ----
 
 Next PR: #216

### DIFF
--- a/curves/src/k256/base_field.rs
+++ b/curves/src/k256/base_field.rs
@@ -373,3 +373,62 @@ impl Fp {
         self.0.to_bytes()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // Tests that compare the inner (unsafe) type with the safe wrapper.
+
+    /// k256::FieldElement.is_zero() panics on unnormalized input.
+    /// In release builds, it would return incorrect results instead of panicking.
+    #[test]
+    #[should_panic(expected = "assertion failed: self.normalized")]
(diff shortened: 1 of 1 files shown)


John-Gong-Math (approved):
LGTM!

iquerejeta (approved):
LGTM

PR #201, opened 2026-02-06
`CircuitField` trait

davidnevadoc:
The main motivation behind `CircuitField` is to resolve tech-debt around field element representation and to allow the integration of curves from other libraries without relying on assumptions about their inner byte layout.

The codebase currently assumes that `PrimeField::Repr` is little-endian. This assumption is embedded in many places, including:

- **BigUint conversions** (`fe_to_big`, `fe_to_bigint` in `utils/util.rs`) -- interpret `to_repr().as_ref()` bytes as little-endian.
- **Bit decomposition** (`fe_to_le_bits`, `le_bits_to_field_elem` in `utils/util.rs`) -- read/write `Repr` bytes assuming LE order.
- **Byte truncation** (`truncate_off_circuit` in `verifier/utils.rs`) -- takes the first N bytes of `to_repr()`, assuming low-order bytes come first.
- **Point encoding** (`JubjubExtended::from_xy` in `ecc/curves.rs`) -- manipulates `to_repr()` byte arrays directly.
- **Tree indexing** (`compute_node_index` in `map/cpu.rs`) -- reads hash output bytes via `to_repr()` and interprets them with `u128::from_le_bytes`.
- **Poseidon hashing** (`poseidon_cpu.rs`) -- serializes field elements via `to_repr()` for hashing and transcript operations.
- **ECDSA signatures** (`utils/ecdsa.rs`) -- decodes r and s from LE byte representations.

These assumptions are fine, since we control the implementation of the curves we use. But it becomes an issue once we start integrating third-party implementations of curves.

## Circuitfield

`CircuitField` extends `PrimeField` with a unified API for integer conversions:

- `to_biguint()` / `from_biguint()` -- canonical BigUint conversions, endianness-agnostic.
- `to_bytes_le()` / `to_bytes_be()` / `from_bytes_le()` / `from_bytes_be()` -- explicit byte-order conversions, with the endianness handling internalized in the implementation.
- `modulus()` -- returns the field modulus as a BigUint. ( Can be generically used, unlike the associated constant in `PrimeField`).

This removes the reliance on implicit assumptions and runtime checks. It should also facilitate future changes or additional constraints on the native field of the circuit, since we now own the trait.

---

## `ScalarExt` associated type in `CircuitCurve`
 Unlike `Base`,  `Scalar` is not defined in the trait CircuitCurve.
 `CircuitCurve` requires `group::Curve`, which requires `group::Group`, which owns this type. In `Group,` we have `type Scalar: PrimeField`  so we cannot further restrict it to be `CircuitField`.
There are 2 options here:
   1. Add the associated type `ScalarExt: CircuitCurve` and use it everywhere instead of `Scalar`.
      Effectively, we want to force `Self::ScalarExt == <Self as Group>::Scalar`, but Rust does not allow this yet (look for issue).
   2. Add an extra trait bound `where CircuitCurve::Scalar: CircuitField`
      This may be the correct way of doing it, but the main drawback is that this clause propagates everywhere and in my opinion, it complicates the codebase unnecessarily.

### DIFF
--- a/circuits/src/instructions/division.rs
+++ b/circuits/src/instructions/division.rs
@@ -202,11 +202,11 @@ pub(crate) mod tests {
         ) -> Result<(), Error> {
             let chip = DivChip::new_from_scratch(&config);
 
-            let x = chip.assign(&mut layouter, self.dividend.clone())?;
+            let x = chip.assign(&mut layouter, self.dividend)?;
             let (q, r) = chip.div_rem(&mut layouter, &x, self.divisor.clone(), None)?;
 
-            chip.assert_equal_to_fixed(&mut layouter, &q, self.expected.0.clone())?;
-            chip.assert_equal_to_fixed(&mut layouter, &r, self.expected.1.clone())?;
+            chip.assert_equal_to_fixed(&mut layouter, &q, self.expected.0)?;
+            chip.assert_equal_to_fixed(&mut layouter, &r, self.expected.1)?;
 
             chip.load_from_scratch(&mut layouter)
         }


circuits/src/circuit_field.rs
```diff
@@ -0,0 +1,276 @@
...
+        let n = BigUint::from_bytes_le(bytes);
+        Self::from_biguint(&n)
+    }
+
+    /// Creates a field element from big-endian bytes.
+    ///
+    /// Returns `None` if the value is not in the canonical range `[0, modulus)`.
+    fn from_bytes_be(bytes: &[u8]) -> Option<Self> {
+        let n = BigUint::from_bytes_be(bytes);
+        Self::from_biguint(&n)
```
miguel-ambrona: I would actually reverse the bytes and call `from_bytes_le`.
Similarly for a default implementation of `to_bytes_be`.

circuits/src/circuit_field.rs
```diff
@@ -0,0 +1,276 @@
...
+                    return None;
+                }
+                let mut padded = [0u8; $repr_size];
+                padded[..bytes.len()].copy_from_slice(&bytes);
+                Self::from_repr(padded.into()).into()
+            }
+
+            fn modulus() -> BigUint {
+                let hex_str = &Self::MODULUS[2..]; // Skip "0x" prefix.
+                BigUint::from_str_radix(hex_str, 16).expect("Invalid modulus hex string")
```
miguel-ambrona: Is it possible to speed this up with some caching? Lazy static?

circuits/src/circuit_field.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Creates a field element from big-endian bytes.
+    ///
+    /// Returns `None` if the value is not in the canonical range `[0, modulus)`.
+    fn from_bytes_be(bytes: &[u8]) -> Option<Self> {
+        let n = BigUint::from_bytes_be(bytes);
+        Self::from_biguint(&n)
+    }
+}
+
+// Macro for implementing CircuitField for LE and BE fields
```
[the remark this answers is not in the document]
miguel-ambrona: Or you meant "MacroS"?

circuits/src/utils/types.rs
```diff
@@ -14,23 +14,23 @@
...
 
-use ff::PrimeField;
 use midnight_proofs::circuit::Value;
 use num_bigint::BigUint;
 #[cfg(any(test, feature = "testing"))]
 use rand::RngCore;
 
 use crate::{
-    types::AssignedNative,
+    field::AssignedNative,
```
miguel-ambrona: Was this intentional?

circuits/src/field/native/native_gadget.rs
```diff
@@ -527,7 +527,7 @@ where
         layouter: &mut impl Layouter<F>,
         value: Value<u8>,
     ) -> Result<AssignedByte<F>, Error> {
-        // We can skip the in-circuit [0, 7]-range-check as this condition will
+        // We can skip the in-circuit [0, 255]-range-check as this condition will
```
miguel-ambrona: 😺

circuits/src/circuit_field.rs
```diff
@@ -0,0 +1,276 @@
...
+///
+/// Implementations must handle endianness internally - callers should not need
+/// to know whether the underlying field uses little-endian or big-endian
+/// representation.
+pub trait CircuitField: PrimeField {
+    /// Byte length of the field representation.
+    const NUM_BYTES: usize;
+
+    /// Fixed-size byte array for the field representation, typically `[u8; NUM_BYTES]`.
+    type Bytes: Copy + Send + Sync + 'static + AsRef<[u8]> + AsMut<[u8]> + Index<usize, Output = u8>;
```
iquerejeta: Can you use `[u8; Self::NUM_BYTES]` instead of `Self::Bytes` and remove this associated type?

circuits/src/ecc/curves.rs
```diff
@@ -27,10 +27,13 @@ use midnight_curves::{
 /// field.
 pub trait CircuitCurve: Curve + Default {
     /// Base field over which the EC is defined.
-    type Base: PrimeField;
+    type Base: CircuitField;
+
+    /// Scalar field with CircuitField bound (same type as Group::Scalar).
+    type ScalarExt: CircuitField;
```
iquerejeta: I think I've already hit this issue - you cannot name this as `Scalar` because a conflict with the associated type of `Group` or `Curve`?

circuits/src/circuit_field.rs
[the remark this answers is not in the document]
davidnevadoc: Meant macros :grin:

davidnevadoc:
>  realized you do not have BE versions for to_bits and from_bits. I think that's ok, but shall we then remove the BE versions of to_bytes and from_bytes?

There is no BE version for bits because we never use it. I think the most common use case the bit-decomposition of a field is the EC point multiplication, when we decompose the scalar. Outside of that it is rarely used. There are some other uses like getting parity / sign of the element, but those are rare and we don't care about BE or LE.
For bytes however, BE an LE are both quite common. It is frequent to find a field element expressed in bytes or limbs in both LE and BE, so it is convenient to have both versions of the `from_bytes_` method. And for `to_bytes_`, we will use the LE version more often but the BE version is very useful for printing the values for debugging or displaying

circuits/src/circuit_field.rs
```diff
@@ -0,0 +1,276 @@
...
+
+            fn to_biguint(&self) -> BigUint {
+                BigUint::from_bytes_le(self.to_repr().as_ref())
+            }
+
+            fn from_biguint(n: &BigUint) -> Option<Self> {
+                let bytes = n.to_bytes_le();
+                if bytes.len() > $repr_size {
+                    return None;
+                }
```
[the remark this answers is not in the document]
davidnevadoc: Exactly!

circuits/src/utils/types.rs
davidnevadoc: nop, part of fixing all the imports, this slipped.

circuits/src/circuit_field.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Converts the field element to little-endian bytes.
+    ///
+    /// The output length is Self::NUM_BYTES.
+    fn to_bytes_le(&self) -> Self::Bytes;
+
+    /// Converts the field element to big-endian bytes.
+    ///
+    /// The output length is Self::NUM_BYTES.
+    fn to_bytes_be(&self) -> Self::Bytes;
```
irakoton: Maybe have a default implementation based on reversing the `le` version? Should be possible even without `Bytes` being defined explicitly as an array type.

circuits/src/ecc/curves.rs
davidnevadoc: Went for `ScalarField`

PR #205, opened 2026-02-11
Add optional names to `VerifierQuery` commitments

miguel-ambrona:
This labeling does not affect the protocol, but can serve for debugging and other purposes (e.g. facilitate the extraction of fixed bases in the verifier_gadget).

### DIFF
--- a/circuits/src/ecc/curves.rs
+++ b/circuits/src/ecc/curves.rs
@@ -169,7 +169,8 @@ impl CircuitCurve for Secp256k1 {
     const NUM_BITS_SUBGROUP: u32 = 256;
 
     fn coordinates(&self) -> Option<(Self::Base, Self::Base)> {
-        Some((self.to_affine().x, self.to_affine().y))
+        let affine = self.to_affine();
+        Some((affine.x, affine.y))
     }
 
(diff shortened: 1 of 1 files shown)


proofs/src/poly/kzg/msm.rs
```diff
@@ -44,20 +46,27 @@ impl<E: Engine> MSMKZG<E> {
...
+            names,
+        }
     }
 
     /// Create a new MSM from a given base (with scalar of 1).
     pub fn from_base(base: &E::G1) -> Self {
         MSMKZG {
             scalars: vec![E::Fr::ONE],
             bases: vec![*base],
+            names: vec![None],
```
iquerejeta: Let's take the name as input.

miguel-ambrona: Why? It will never be called with a name.

iquerejeta: It is not clear to me from the interface. We can add it if we ever come up with a scenario where we need that.

iquerejeta (approved):
Looks good to me 👍

davidnevadoc (approved):
LGTM!

PR #214, opened 2026-02-16
Small fixes in `CircuitField`

davidnevadoc:
- Fix from_bits_le when receiving small number of bits (less than Self::NUM_BITS).
- Fix to_bits_le for F::ZERO.
- Add tests


miguel-ambrona (approved):
Great!

Can you also mention that "from_bytes" requires exactly `F::NUM_BYTES` bytes?

PR #225, opened 2026-02-23
Fix MSM identity handling

davidnevadoc:
Fix `msm_best`:
 - Incorrect handling of identity points in the input resulted in a panic.
 - Incorrect `contain` in schedule caused it to return true for 0 bucket when empty.
 
 The verifier does not use this version of the MSM so it was not affected by the bug. (It uses the native bltrs version).

### DIFF
--- a/curves/src/msm.rs
+++ b/curves/src/msm.rs
@@ -143,6 +143,9 @@ fn batch_add<C: CurveAffine>(
     }
 }
 
+/// Wrapper to provide direct access to affine coordinates.
+/// This avoids repeated calls to `CurveAffine::coordinates()` (which returns
+/// a `CtOption` and performs an identity check).
 #[derive(Debug, Clone, Copy)]
 struct Affine<C: CurveAffine> {
     x: C::Base,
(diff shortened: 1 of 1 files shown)


curves/src/msm.rs
```diff
@@ -467,10 +470,13 @@ pub fn msm_best<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Curve {
         return msm_parallel(coeffs, bases);
     }
 
-    // coeffs to byte representation
-    let coeffs: Vec<_> = coeffs.par_iter().map(|a| a.to_repr()).collect();
-    // copy bases into `Affine` to skip in on curve check for every access
-    let bases_local: Vec<_> = bases.par_iter().map(Affine::from).collect();
+    // Filter out identities. Transform scalars to bytes and bases to affine.
```
miguel-ambrona:
```suggestion
    // Filter out identity points. Transform scalars to bytes and bases to affine.
```

curves/src/msm.rs
```diff
@@ -493,8 +499,7 @@ pub fn msm_best<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Curve {
 
                 if sched.contains(buck_idx) {
                     // greedy accumulation
-                    // we use original bases here
-                    j_bucks[buck_idx].add_assign(&bases[base_idx], sign);
+                    j_bucks[buck_idx].add_assign(&bases_local[base_idx].eval(), sign);
```
miguel-ambrona: What's the effect of this change?

curves/src/msm.rs
```diff
@@ -467,10 +470,13 @@ pub fn msm_best<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Curve {
...
-    // coeffs to byte representation
-    let coeffs: Vec<_> = coeffs.par_iter().map(|a| a.to_repr()).collect();
-    // copy bases into `Affine` to skip in on curve check for every access
-    let bases_local: Vec<_> = bases.par_iter().map(Affine::from).collect();
+    // Filter out identities. Transform scalars to bytes and bases to affine.
+    let (coeffs, bases_local): (Vec<_>, Vec<_>) = coeffs
+        .par_iter()
+        .zip(bases.par_iter())
+        .filter(|(_, b)| !bool::from(b.is_identity()))
+        .map(|(c, b)| (c.to_repr(), Affine::from(b)))
```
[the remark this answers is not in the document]
davidnevadoc: `Affine::from()` is not normalizing. It is turning affine points into the `Affine` struct used by the MSM. It is not very elegant TBH...

curves/src/msm.rs
davidnevadoc: We need to used the filtered out bases to guarantee the index is right.

PR #229, opened 2026-02-26
First wave of parsing tools (1/4)

irakoton:
# Substring checks

This PR is the first part (excluding the refactor) of several features introduced for parsing passports in-circuit, in particular ASN.1 credentials. Follows #240 

This PR introduces substring checks to the Scanner chip.

## Lookup argument

Substring checks implement a function, `check_bytes`, that asserts that a sequence of `AssignedByte<F>`, `sub`, is a subsequence of `sequence` starting from index `idx : AssignedNative<F>`. This is implemented through dynamic lookups, intuitively to prove that `queries ⊆ table`, where

```
queries = { (sub[0], idx), (sub[1], idx+1), ..., (sub[p], idx+p) }
table = { (sequence[0], 0), (sequence[1], 1), ..., (sequence[n], n) }
```

To save columns, instead of looking up pairs `(byte, index)`, we range check `idx` and look up `index * 257 + byte`. The bound `257` comes from the fact that `byte` will sometimes take the dummy value `256` for padding purposes in the following PRs. The concrete lookup argument uses 4 columns in total; at row `i`:

1. 1 advice column to store `sequence[i]`
2. 1 advice column to store `(idx + i) * 257 + sub[i]`, computed using `native_gadget.linear_combination`
3. 1 fixed column to store `i`
4. 1 fixed column to store a fresh, non zero tag (used for isolation between calls and with the rest of `zk_stdlib`).

The `sub` bytes are already "indexed/packed" when assigned, whereas `sequence` bytes are packed by the lookup `Expression`. This choice comes from the fact that it's cheaper to assign the constant indexes `0,1,...,n` in a fixed column.

## Caching

In practice, the function `check_byte` itself only records the function call in a global reference, while the circuit logic is truly computed when calling `load` at the end of circuit synthesis. Through this, substring checks get an implicit caching mechanism, making it so that checking that `sub1`, `sub2`,... are substrings of `sequence` is in practice represented by a single lookup.

## Performances

On the Credential Property verification, using this parsing technique instead of the previous manual index computation reduces the number of rows in the goldenfiles from ~52k to ~22k.

### DIFF
--- a/circuits/src/parsing/scanner/mod.rs
+++ b/circuits/src/parsing/scanner/mod.rs
@@ -332,9 +332,13 @@ where
         // The lookup then checks: (tag, packed_query) ∈ {(tag, table_packed)}.
         //
         // When sel=OFF, both sides reduce to (tag, query), i.e., a tautology, so rows
-        // not used by substring checks are unconstrained. The tag fixed column will
-        // default to 0 outside of the substring check regions, isolating the tables
-        // from the unrelated regions.
+        // not used by substring checks are unconstrained.
+        //
+        // Invariant: the tag column is 0 on every row that is not part of a substring
+        // check region, and non-zero (a unique positive integer) inside each region.
+        // This isolates independent substring checks from each other and from
+        // unrelated rows: a query tagged T can only match table entries with the
+        // same tag T, and rows with tag 0 never participate in any lookup.
         meta.lookup_any("substring lookup", |meta| {
             let sel = meta.query_selector(q_substring);
             let not_sel = Expression::Constant(F::ONE) - sel.clone();
--- a/circuits/src/parsing/scanner/substring.rs
+++ b/circuits/src/parsing/scanner/substring.rs
@@ -75,7 +75,10 @@
 //! ```
 //!
 //! Tags and the table index column are written in fixed columns; the
-//! remaining columns (table bytes and query entries) are advice columns.
+//! remaining columns (table bytes and query entries) are advice columns. The
+//! invariant is that the tag is >0 in substring-check regions, and 0 in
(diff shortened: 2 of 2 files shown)


circuits/src/parsing/scanner/mod.rs
```diff
@@ -232,16 +274,50 @@ where
...
+            let not_sel = Expression::Constant(F::ONE) - sel.clone();
+            let index = meta.query_fixed(*index_col, Rotation::cur());
+            let tag = meta.query_fixed(tag_col, Rotation::cur());
+            let shift = Expression::Constant(F::from(ALPHABET_MAX_SIZE as u64 + 1));
+
+            let table = meta.query_advice(state_col, Rotation::cur());
+            let query = meta.query_advice(letter_col, Rotation::cur());
+
+            vec![
+                (tag.clone(), sel.clone() * tag.clone()),
```
iquerejeta: What are we doing here? Isn't this lookup always going to be satisfied?

circuits/src/parsing/scanner/automaton_chip.rs
```diff
@@ -122,7 +122,83 @@ where
...
+                    for state in automaton.1.final_states.iter() {
+                        add_entry(*state, F::from(ALPHABET_MAX_SIZE as u64), F::ZERO, F::ZERO)?
+                    }
+                }
+                Ok(())
+            },
+        )
+    }
+}
+impl<LibIndex, F> ScannerChip<LibIndex, F>
```
iquerejeta:
```suggestion
}

impl<LibIndex, F> ScannerChip<LibIndex, F>
```

PR #233, opened 2026-02-26
First wave of parsing tools (3/4)

irakoton:
# Parser for plaintext Passport Data

This PR is the third part of several features introduced for parsing passports in-circuit, in particular ASN.1 credentials. Follows #232 

Adds the DG1 (Data Group 1) spec in the automaton static library. This is the 88-byte long identification string that typically appears on the first pages of biometric passports, and contains the main data to extract from the identity document (e.g., DoB).

Nothing technical to mention, it's just a translation (sources in doc comments) of an official spec. A few simpler fields more related to ASN.1 will be added in the second wave of parsing tools.

### DIFF
--- a/circuits/src/parsing/scanner/static_specs.rs
+++ b/circuits/src/parsing/scanner/static_specs.rs
@@ -147,24 +147,24 @@ pub enum StdLibParser {
     ///
     /// In the MRZ format, only uppercase letters, digits, and `<` (representing
     /// special characters such as spaces or dashes, as well as padding and
-    /// separators), are used. The following fields are then marked as follows;
-    /// note that checksum fields are not mentioned, as they are neither marked
-    /// nor verified by the parser.
+    /// separators), are used. The following fields are then output as follows;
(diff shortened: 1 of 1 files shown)


circuits/src/parsing/scanner/static_specs.rs
```diff
@@ -280,6 +348,79 @@ fn spec_jwt() -> Regex {
...
+        passport_type,
+        Regex::uppercase_letter().mark(&|_| Some(2)).repeat(3),
+        name_block(3).separated_non_empty_list("<".into()),
+    ]);
+    // Given names in the first line of the DG1, prefixed with the separator. The
+    // separators `<` are not marked.
+    let given_names = Regex::cat([
+        "<<".into(),
+        name_block(4).separated_non_empty_list("<".into()),
+    ]);
```
iquerejeta: What happens if there is more than two occurrences of the 'non_empty_list'? can that happen?

PR #234, opened 2026-02-26
First wave of parsing tools (4/4)

irakoton:
# Parallel Lookups in Scanner Chip

This PR is the fourth part of several features introduced for parsing passports in-circuit, in particular ASN.1 credentials. Follows #233 

Restructures the chips of the `ScannerChip` to include parametrisable parallel lookups. There are two constants, `SUBSTRING_PARALLELISM` and `AUTOMATON_PARALLELISM`, respectively controlling the level of parallelism of substring checks and automaton checks. This is, in short, the number of parallel lookups used by the sub-chips. 

## Substring checks

As a reminder, non-parallelised substring checks operate on 4 columns:

1. 1 "tag" fixed column for call isolation
2. 1 "index" fixed column for indexing the `sequence` bytes
3. 1 "packed" advice column containing the `sub` bytes batched with their index
4. 1 "sequence" advice column containing the `sequence` bytes.

By using parallel lookups, we may manage several of substring checks in parallel (each call keeps the same cost, but several calls are managed on the same rows). The two fixed columns can be shared between the parallel calls, so each parallel call costs `NB_SUBSTRING_COLS = 2` advice columns.

In practice, the calls are grouped by batches of `SUBSTRING_PARALLELISM`, and padded so that they all consume the same number of rows. The `sequence` arguments are padded with `256` (invalid byte), and `sub` arguments are padded by stuttering on their last "packed" value.

## Automaton parsing

Each parallel lookup requires 3 advice columns.

Simply parses `AUTOMATON_PARALLELISM` bytes by row instead of 1, when calling `ScannerChip::parse`.

## Recommended parallelism degree

In the current state of the code, `SUBSTRING_PARALLELISM = 3` and `AUTOMATON_PARALLELISM = 2`, for a total of 6 advice columns in the chip. Keeping in mind that parsers are mostly needed for identity-related purposes, they often involve RSA signatures that require 8 columns anyway. So it's likely reasonable to have `AUTOMATON_PARALLELISM = 2~3`, and `SUBSTRING_PARALLELISM = 4` (for a total of 8~9 advice columns).

## Performance improvements

Any circuit region performing automaton parsing has its number of rows divided by `AUTOMATON_PARALLELISM`. Substring checks are not more efficient per se, but make it so that adding calls with other `sequence` arguments becomes free, up to `SUBSTRING_PARALLELISM - 1` times. Current identity examples only involve 1 unique `sequence` argument, but full ASN.1 parsers are likely to involve more.

## ⚠️ `main` branch maintenance ⚠️

Of course, increasing the number of parallel lookups mostly makes sense in `main` when LogUp is available. Also, they are here performed with calls to `meta.lookup*` in `for` loops, but would benefit from `batch_lookup*`, only available in `main`.


circuits/src/parsing/scanner/automaton_chip.rs
```diff
@@ -77,23 +89,28 @@ where
...
+                        self.apply_one_transition(
+                            &mut region,
+                            automaton,
+                            &mut state,
+                            letter,
+                            batch,
+                            &mut markers,
+                            &mut offset,
+                        )?;
+                    }
```
iquerejeta: Rather than adding `batch` to the inputs, have you though of using `chunks_exact`? Might be a cleaner version, and then you handle the remaining values specifically.

Issue #231, opened 2026-02-26
Reduce public-input representation for foreign-field elements by packing limbs


miguel-ambrona:
Currently, [`constrain_as_public_input`](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/field/foreign/field_chip.rs#L490) exposes every (normalized) limb of an emulated field element as a separate public input (PI).

However, multiple limbs can be packed into a single native-field element via a linear combination with powers of the base (e.g. `packed = limb[0] + BASE * limb[1] + BASE² * limb[2] + ...`) as long as the result is guaranteed to fit without wrapping.

### What needs to be done

1. Implement in-circuit packing in [`constrain_as_public_input`](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/field/foreign/field_chip.rs#L490) with an in-circuit linear combination.

2. Update [`as_public_input`](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/field/foreign/field_chip.rs#L481) accordingly.

3. Also, [`assign_as_public_input`](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/field/foreign/field_chip.rs#L500).

4. And do not forget about the off-circuit counterpart: [`Instantiable::as_public_input`](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/field/foreign/field_chip.rs#L130).

### Motivation

The main benefit is in recursive (IVC) circuits, where public inputs are hashed in-circuit with Poseidon. Fewer PIs means fewer Poseidon absorptions, which directly reduces the constraint count of the recursive verifier. For a circuit with many foreign-field public inputs this can be a significant saving.

Evaluate the performance gain.

### Scope

This targets the `ForeignFieldInstructions` public-input methods in [`field_chip.rs`](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/field/foreign/field_chip.rs). The underlying `FieldEmulationParams` trait and the foreign-field arithmetic gates should not be affected.

PR #235, opened 2026-02-27
Batch foreign-field limbs when constraining as public inputs

miguel-ambrona:
Addressing https://github.com/midnightntwrk/midnight-zk/issues/231.

The caveat is that we can no longer have an optimized implementation of `assign_as_public_input` and that `constrain_as_public_input` takes a few rows (to perform the linear combination in-circuit).

Despite this, the overall improvement is significant. Note that assignment of foreign field elements as public inputs is a rare operation. However, with this change, the IVC circuit (over BLS12-381)  improves about 7300 rows. That's about 3.25%, not bad for such a simple change.


circuits/src/field/foreign/field_chip.rs
```diff
@@ -502,22 +516,11 @@ where
...
-        let assigned_limbs = self.native_gadget.assign_many(layouter, &limbs)?;
-        let assigned_field = AssignedField::<F, K, P> {
-            limb_values: assigned_limbs,
-            limb_bounds: well_formed_bounds::<F, K, P>(),
-            _marker: PhantomData,
-        };
-        self.constrain_as_public_input(layouter, &assigned_field)?;
-        Ok(assigned_field)
+        // Do NOT optimize this implementation. Since we batch various limbs when
+        // constraing as public inputs, we cannot skip the range-checks on the limbs.
```
iquerejeta: I don't see range-checks here

iquerejeta (approved):
LGTM

irakoton (approved):
LGTM/

PR #239, opened 2026-03-01
ci: update scanner

gilescope:
Use latest versions of security scanners. scan is now not a sub-action.


adamreynolds-io (approved):

iquerejeta (approved):

.github/workflows/scan.yaml
gilescope:
```suggestion
    branches: [main, next]
```

PR #241, opened 2026-03-02
Foreign Edwards chip (1/2)

remalue:
This PR introduces foreign arithmetic over twisted Edwards curves. It is `Part 1` in a series of 2 PRs.

`Part 1`: this PR
`Part 2`: #242

### DIFF
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -32,4 +32,4 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  #v6.0.2
 
       - name: Scan code
-        uses: midnightntwrk/upload-sarif-github-action@9856edc26a43e2d0cc3b391888bce2295377bdd7
+        uses: midnightntwrk/upload-sarif-github-action@07dad711370cc5985885ebcf07cb8c9264bc4167


circuits/src/instructions/ecc.rs
```diff
@@ -309,10 +309,10 @@ pub(crate) mod tests {
             _marker: PhantomData,
         };
         let log2_nb_rows = match operation {
-            Operation::Msm => 17,
-            Operation::MsmBounded => 16,
-            Operation::MulByConstant => 16,
-            _ => 10,
+            Operation::Msm => 19,
```
davidnevadoc: Did the `k` increase because the foreign Edwards ops are more expensive?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -0,0 +1,1172 @@
...
+            // The doubling in the last iteration is not needed
+            if i + 1 < scalar_bits.len() {
+                p = self.double(layouter, &p)?;
+            }
+        }
+
+        Ok(res)
+    }
+
+    fn point_from_coordinates(
```
davidnevadoc: Maybe we need an `unsafe` version of this. The scenario I imagine is the same one in which we need it for the native curve: 
A map-to-curve algorithm that gives us coordinates on the emulated curve, not necessarily in the subgroup. 
We would need a method for `coordinates -> point` without the subgroup check, and then clear the cofactor.

circuits/src/ecc/foreign/weierstrass_chip.rs
```diff
@@ -408,9 +409,9 @@ where
 /// Inherit assignment instructions for [AssignedField], from the
 /// `scalar_field_chip` when the emulated field field is the scalar field.
 /// Mind the binding `S: ScalarFieldInstructions<F, Scalar = AssignedField<F,
-/// C::ScalarField>>` of this implementation.
+/// C::Scalar>>` of this implementation.
```
miguel-ambrona: Is this correct? Shouldn't it remain `C::ScalarField`?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -799,3 +804,369 @@ where
...
+    ecc_tests!(test_msm);
+    ecc_tests!(test_msm_by_bounded_scalars);
+    ecc_tests!(test_mul_by_constant);
+    ecc_tests!(test_coordinates_edwards);
+
+    #[test]
+    fn test_assert_in_subgroup() {
+        run_test_assert_in_subgroup::<Curve25519>();
+        run_test_assert_in_subgroup::<JubjubExtended>();
+    }
```
miguel-ambrona: Maybe these tests should be made generic.

Shall we unify the way we apply subgroup checks. Through the `EccInstructions` trait?

circuits/src/instructions/ecc.rs
remalue: Yes, this increased because of Edwards ops. I matched a few more operations (and could reduce a few by `-1` - maybe an artefact from cleanup), but doesn't change the overall increase.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -0,0 +1,1172 @@
...
+}
+
+impl<F, C, B> Instantiable<F> for AssignedForeignEdwardsPoint<F, C, B>
+where
+    F: CircuitField,
+    C: EdwardsCurve,
+    B: FieldEmulationParams<F, C::Base>,
+{
+    // TODO: why default to 0?
```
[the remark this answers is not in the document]
remalue: I'd leave as is for now (unwrap to `(0,0)`), since also in Weierstrass form it's not unwrapped to the identity.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -0,0 +1,1172 @@
...
+    ) -> Result<Self::Point, Error> {
+        let scalars = scalars
+            .iter()
+            .map(|s| (s.clone(), C::ScalarField::NUM_BITS as usize))
+            .collect::<Vec<_>>();
+
+        self.msm_by_bounded_scalars(layouter, &scalars, bases)
+    }
+
+    fn msm_by_bounded_scalars(
```
davidnevadoc: This implementation is fine for this PR. 
But we could add a note here that there are several improvements available:
 - Batching equal points
 - Filtering scalars
 - Using windowed method

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -11,31 +11,44 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! TODO: Doc the chip.
+//! Elliptic curve (in twisted Edwards form) operations over foreign fields.
+//! This module supports curves of the form a*x^2 + y^2 = 1 + d*x^2*y^2, where
+//! a is square and d is non-square.
```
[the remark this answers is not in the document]
remalue: Done.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -186,6 +281,60 @@ where
...
+impl<F, C, B, S, N> ForeignEdwardsEccChip<F, C, B, S, N>
+where
+    F: CircuitField,
+    C: EdwardsCurve,
+    B: FieldEmulationParams<F, C::Base>,
+    S: ScalarFieldInstructions<F>,
+    S::Scalar: InnerValue<Element = C::ScalarField>,
+    N: NativeInstructions<F>,
+{
+    /// Converts a subgroup point to [AssignedForeignEdwardsPoint].
```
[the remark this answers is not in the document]
remalue: It's the same wording as in the Weierstrass chip. Any suggestion?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -0,0 +1,1172 @@
...
+
+        let base_chip = self.base_field_chip();
+
+        let r_value = p.value().zip(q.value()).map(|(p, q)| p + q);
+        let r = self.assign_point_unchecked(layouter, r_value)?;
+
+        let px_qy = base_chip.mul(layouter, &p.x, &q.y, None)?;
+        let py_qx = base_chip.mul(layouter, &p.y, &q.x, None)?;
+        let py_qy = base_chip.mul(layouter, &p.y, &q.y, None)?;
+        let px_qx = base_chip.mul(layouter, &p.x, &q.x, None)?;
```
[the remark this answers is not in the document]
davidnevadoc: I think this can be a fairly easy optimization so it could be included in this PR ( and it may lower the higher K issue mentioned above). But it is also fine to open an issue and leave it as a future improvement.

Issue #245, opened 2026-03-03
Get rid of `constant_commitment` in the `PolynomialCommitmentScheme` trait


miguel-ambrona:
As requested [here](https://github.com/midnightntwrk/midnight-zk/pull/190#discussion_r2845990409).

miguel-ambrona:
Closed by https://github.com/midnightntwrk/midnight-zk/pull/246

PR #246, opened 2026-03-03
Separate constant part of linear poly

davidnevadoc:
Split the linearization polynomial into its homogeneous and its constant parts:
`L(X) = L'(X) + C`.

The query now checks that `L'` evaluates to `-C` on `x` instead of checking that `L` evaluates to `0` on `x`.

Fixes https://github.com/midnightntwrk/midnight-zk/issues/245.

### DIFF
--- a/circuits/src/verifier/verifier_gadget.rs
+++ b/circuits/src/verifier/verifier_gadget.rs
@@ -411,8 +411,8 @@ impl<S: SelfEmulation> VerifierGadget<S> {
     ///        to the affine term `C` of the linearization polynomial),
     /// * `h_k` are commitments to the limbs of the quotient polynomial.
     ///
-    /// The linearization polynomial is split into its homogeneous and affine
-    /// parts: `L(X) = L'(X) + C`. Both parts are returned separately.
+    /// The linearization polynomial is split into its non-constant and constant
+    /// parts: `L(T) = L'(T) + C`. Both parts are returned separately.
     ///
     /// # Arguments
     ///
@@ -424,8 +424,8 @@ impl<S: SelfEmulation> VerifierGadget<S> {
     ///
     /// # Returns
     ///
-    /// A tuple `(L'(X), C)`, where `L'(X)` is an [AssignedMsm] and `C` is a
-    /// scalar. The verifier uses `-C` as the expected evaluation of `L'(X)` at
+    /// A tuple `(L'(T), C)`, where `L'(T)` is an [AssignedMsm] and `C` is a
+    /// scalar. The verifier uses `-C` as the expected evaluation of `L'(T)` at
     /// `x`.
     #[allow(clippy::too_many_arguments)]
     #[allow(clippy::type_complexity)]
@@ -473,7 +473,7 @@ impl<S: SelfEmulation> VerifierGadget<S> {
         // Fully evaluated identities (None) are excluded from the MSM.
         // Their accumulated scalar C is returned so the caller can use -C as
         // the expected evaluation of L'(X) at x.
-        let mut lin_poly_affine_term = zero;
+        let mut lin_poly_constant_term = zero;
         for (col_idx, eval) in grouped_points {
             match col_idx.map(|column_index| vk.fixed_commitment_name(column_index)) {
                 Some(com) => {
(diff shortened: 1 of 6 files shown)


proofs/src/plonk/linearization/prover.rs
```diff
@@ -33,26 +36,30 @@ use crate::{
...
     splitting_factor: F,
     quotient_limbs: Vec<Polynomial<F, Coeff>>,
-) -> Polynomial<F, Coeff> {
+) -> (Polynomial<F, Coeff>, F) {
     let mut y_pow = F::ONE;
+    let mut lin_poly_affine_term = F::ZERO;
     let lin_poly = expressions.iter().rev().fold(
         Polynomial::init(pk.vk.get_domain().n as usize),
-        |mut acc, (col_idx, eval)| match col_idx {
+        |acc, (col_idx, eval)| match col_idx {
```
remalue, line 53: Why remove `mut`?

proofs/src/plonk/linearization/prover.rs
```diff
@@ -20,10 +20,13 @@ use crate::{
...
 ///        multiplicative selector, or,
 ///      - (ii) 1 (in case the corresponding identity `id_j` has been fully
-///        evaluated and, thus, the resulting scalar `id_j(x)` is part of the
-///        constant term of the linearization polynomial),
+///        evaluated and, thus, the resulting scalar `id_j(x)` contributes to
+///        the affine term `C` of the linearization polynomial),
 /// * `h_k(T)` are the limbs of the quotient polynomial.
 ///
+/// The linearization polynomial is split into its homogeneous and affine parts:
+/// `L(X) = L'(X) + C`. Both parts are returned separately.
```
remalue: Use T as indeterminate.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -409,12 +406,14 @@ impl<S: SelfEmulation> VerifierGadget<S> {
...
-    ///      - (ii) the commitment to the constant polynomial `P(X) = 1` (in
-    ///        case the corresponding identity `id_j` has been fully evaluated
-    ///        and, thus, the resulting scalar `id_j(x)` is part of the constant
-    ///        term of the linearization poly),
+    ///      - (ii) 1 (in case the corresponding identity `id_j` has been fully
+    ///        evaluated and, thus, the resulting scalar `id_j(x)` contributes to
+    ///        the affine term `C` of the linearization polynomial),
     /// * `h_k` are commitments to the limbs of the quotient polynomial.
     ///
+    /// The linearization polynomial is split into its homogeneous and affine parts:
```
remalue: Why not simply non-constant and constant part? The non-constant part (as a univariate poly) is, in general, not homogeneous.

Also later, in the dbg_assert, constant part is used. Better to be consistent here.

proofs/src/plonk/bench/prover.rs
```diff
@@ -574,9 +574,9 @@ where
     };
 
     debug_assert_eq!(
-        eval_polynomial(&linearization_poly, x),
-        F::ZERO,
-        "The linearization poly should evaluate to zero at the evaluation challenge x."
+        eval_polynomial(&homogeneous_lin_poly, x),
+        -lin_poly_affine_term,
+        "L'(x) should equal -C, where C is the constant part from fully evaluated identities."
```
remalue: "where C is the constant part from fully evaluated identities" -> "where C is the constant part of the linearization polynomial/`L'`"

circuits/src/verifier/verifier_gadget.rs
miguel-ambrona: I figured a technical work like "homogeneous" was very precise and also rare across this library, so it was appropriate for describing this new concept.

I'm not very fond of "non-constant". "Variable" is also a bit ambiguous in our context.

proofs/src/plonk/linearization/prover.rs
miguel-ambrona: Because it is not being mutated anymore. The `None` case is handled differently now.

remalue: And why not mutate the `Some` case?

circuits/src/verifier/verifier_gadget.rs
remalue: Hmm...I am not convinced by that. "Homogeneous" has a clear meaning in the context of polynomials, and the non-constant part (as univariate poly) doesn't satisfy this meaning.

proofs/src/plonk/linearization/prover.rs
miguel-ambrona: I'll ask David to explore that in his optimizations PRs.

PR #256, opened 2026-03-10
More optimizations on ECC

miguel-ambrona:
- Minor fix: Sort point sets deterministically to ensure consistency between off-circuit and in-circuit verifier. Also, make sure the set with a single point (x) is the first. This is necessary because it is the set involving fixed queries. This is the only set that we do not internally collapse, being the first allows us to give it a scaling factor of 1 (the first x4 power) which does not spoil the truncated challenges even if we are not collapsing it.

- Filter out fixed identity points from MSMs.

- Move advice queries before instance queries. This is useful in the verifier gadget if an instance commitment is hard-coded to be the identity (unused). If that is the case, we do not want the instance commitment to be the first in the MSM (where it would enjoy a scalar of 1), we want a non-trivial point to enjoy such scalar of 1 whereas the instance commitment can be optimized away elsewhere thanks to the previous optimization.


davidnevadoc (approved):
LGTM!

PR #259, opened 2026-03-10
Optimize multiplication in `BigUint_gadget`

davidnevadoc:
- Add dedicated `square()` that exploits symmetry (only n(n+1)/2 products instead of n^2).
  - Batch product accumulation in both `mul` and `square`: collect all products per output limb position and accumulate with a  single `linear_combination` instead of per-product native `add`.
  - Short-circuit mul(x, x) to square(x) if the input cells are equal.

  ## Impact
  - `mul`: 497 → 415 rows (−16.5%)
  - `square` : 347 rows (−30% vs original mul)
  - RSA signature example (zkir): 83820 → 61.688 rows (-26.4%)

### DIFF
--- a/zk_stdlib/CHANGELOG.md
+++ b/zk_stdlib/CHANGELOG.md
@@ -15,6 +15,8 @@ verification keys break backwards compatibility.
 * Added an example verifying an ethereum signature [#177](https://github.com/midnightntwrk/midnight-zk/pull/177)
 * `setup_vk_with_k` for generating a verifying key with an explicit circuit size parameter [#227](https://github.com/midnightntwrk/midnight-zk/pull/227)
 * Expose `verifier_gadget` and `bls12_381_scalar` (native gadget) from `ZkStdLib` [#227](https://github.com/midnightntwrk/midnight-zk/pull/227)
+* Add `square`, `mod_square` operations in `BigUintGadget` [#259](https://github.com/midnightntwrk/midnight-zk/pull/259)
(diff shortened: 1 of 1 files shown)


circuits/src/biguint/biguint_gadget.rs
```diff
@@ -430,6 +430,48 @@ where
...
+            // Off-diagonal terms: 2 * x_i * x_j for j > i.
+            for j in (i + 1)..x.limbs.len() {
+                let p =
+                    native_gadget.mul(layouter, &x.limbs[i], &x.limbs[j], Some(F::from(2u64)))?;
+                let p_bound = 1 + x.limb_size_bounds[i] + x.limb_size_bounds[j];
+                limbs[i + j] = native_gadget.add(layouter, &limbs[i + j], &p)?;
+                limb_size_bounds[i + j] =
+                    bound_of_addition(limb_size_bounds[i + j], p_bound);
+            }
+        }
```
miguel-ambrona: Very nice!

This can be optimized even further if instead of calling `native_gadget.add` as soon as you need it, you build a pool of terms to be added for every limb and you call `native_gadget.linear_combination` once at the end (we can add more than one term at a time).

Another alternative to explore is to use `add_and_mul` directly.
So with a single row you compute `p` and add it to the corresponding limb.

miguel-ambrona (approved):
Fantastic improvements! 🚀

PR #277, opened 2026-03-18
feat(msm): add MSMCompletenessFluke error and proactive checks

### LINKED ISSUE #262: Add a dedicated `plonk::Error` for MSM completeness failures caused by bad randomness `r`
## Context                                                                                                                               

The windowed MSM implementation in `circuits/.../ecc_chip.rs` uses a randomly sampled point `r` as an accumulator offset to enable incomplete addition throughout the double-and-add loop. Point `r` is sampled uniformly ([`circuits/src/ecc/foreign/ecc_chip.rs:1928`](https://github.com/midnightntwrk/midnight-zk/blob/next/circuits/src/ecc/foreign/ecc_chip.rs#L1928)) by the prover.

## The Problem

For proof generation to succeed, the sampled `r` must not cause any of the following "bad events" during the loop:
 - The accumulator collides with an addend (x-coordinate collision), causing `incomplete_assert_different_x` to fail.
 - Any intermediate accumulator value becomes the identity.
 - The table construction loop hits an x-coordinate collision or produces the identity.
  
Each of these events has negligible probability over the random choice of `r`. However, if the bad event occurs, proof generation currently fails with a generic, hard-to-diagnose constraint error, giving the caller no way to distinguish a completeness fluke from a real bug.
                                                                                                                                        
## Proposed Fix

Introduce a new plonk::Error variant that is returned specifically when one of the above conditions is detected during witness generation. Such error can be more general, to hint that the proof generation failed for a "bad luck" reason, not a fundamental one in the proof system. The caller can then catch this error and try their luck again, rather than treating it as a fatal failure.

This is not strictly necessary, given that we will probably never see this error being triggered, but it is good practice to fail gracefully and informatively rather than producing a cryptic constraint system error that could be mistaken for a soundness bug.

Kanasjnr:
# Overview

This PR introduces a dedicated `plonk::Error` variant, `MSMCompletenessFluke`, and integrates proactive checks in the windowed MSM implementation to detect rare failures caused by unfavorable random sampling.

The MSM implementation uses a randomly sampled point `r` as an accumulator offset to enable incomplete addition. In extremely rare cases, `r` can cause x-coordinate collisions or result in the identity point, leading to synthesis failures. These events now return the specific `MSMCompletenessFluke` error instead of generic constraint failures, allowing provers to identify the cause and retry with different randomness.

## Submission Checklist

- [x] Useful pull request description
- [x] Tests are provided (see [test_msm_completeness_fluke](cci:1://file:///Users/kanas/Desktop/opensource/midnight-zk/circuits/src/ecc/foreign/ecc_chip.rs:2356:4-2449:5) in [ecc_chip.rs](cci:7://file:///Users/kanas/Desktop/opensource/midnight-zk/circuits/src/ecc/foreign/ecc_chip.rs:0:0-0:0))
- [x] Key commits have useful messages
- [x] All check jobs of the CI have succeeded (`cargo test` and `cargo clippy` passed locally)
- [x] Self-reviewed the diff
- [x] Reviewer requested
- [ ] Update README.md file (Not applicable)
- [ ] Update documentation (Not applicable)
- [x] No new todos introduced

## Links

- Closes #262

### DIFF
--- a/circuits/CHANGELOG.md
+++ b/circuits/CHANGELOG.md
@@ -27,6 +27,7 @@ verification keys break backwards compatibility.
 * Implementation of substring checks [#229](https://github.com/midnightntwrk/midnight-zk/pull/229)
 * Foreign arithmetic over twisted Edwards curves [#241](https://github.com/midnightntwrk/midnight-zk/pull/241)
 * Custom gate for foreign point addition over twisted Edwards curves [#242](https://github.com/midnightntwrk/midnight-zk/pull/242)
+* add MSMCompletenessFluke error [#277](https://github.com/midnightntwrk/midnight-zk/pull/277)
 
 ### Changed
 * Filter out compile-time identity points in MSM [#256](https://github.com/midnightntwrk/midnight-zk/pull/256)


circuits/src/ecc/foreign/ecc_chip.rs
```diff
@@ -2000,8 +2036,23 @@ where
...
+                            let addx = (*addv).into().coordinates().unwrap().0;
+                            ax == addx
+                        }
+                    })
+                    .map_err(|_| Error::CompletenessFailure)?;
                 self.incomplete_assert_different_x(layouter, &acc, &addend)?;
                 acc = self.incomplete_add(layouter, &acc, &addend)?;
+                acc.value()
+                    .error_if_known_and(|p| C::CryptographicGroup::is_identity(p).into())
+                    .map_err(|_| Error::CompletenessFailure)?;
```
miguel-ambrona: Remove. Function `incomplete_add` can never return the identity if its preconditions are satisfied.

Kanasjnr:
> @Kanasjnr Please, sign your commits.
> 
> You may use the occasion to bring back this comment that should not have been removed:
> 
> ```
> // Assert that α.x ≠ p.x (note that α and -α have the same x-coordinate).
> ```

Done!!

miguel-ambrona:
> > @Kanasjnr Please, sign your commits.
> > You may use the occasion to bring back this comment that should not have been removed:
> > ```
> > // Assert that α.x ≠ p.x (note that α and -α have the same x-coordinate).
> > ```
> 
> Done!!

All commits must be signed.

Kanasjnr:
> > > @Kanasjnr Please, sign your commits.
> > > You may use the occasion to bring back this comment that should not have been removed:
> > > ```
> > > // Assert that α.x ≠ p.x (note that α and -α have the same x-coordinate).
> > > ```
> > 
> > 
> > Done!!
> 
> All commits must be signed.

should be done now check

Issue #278, opened 2026-03-19
Fix `proof_size` in `circuit_model`


miguel-ambrona:
The proof size reported by the `circuit_model` is no longer accurate in the `main` branch.

The mismatch may be due to recent changes like linearization or logup.

I suggest we add a CI check for the circuit_model proof size of various circuits, comparing it to the actual size of a generated proof. That will force us keep this function up to date if the protocol changes.

PR #280, opened 2026-03-20
Sync cost model proof `size` with the protocol

miguel-ambrona:
Fixes https://github.com/midnightntwrk/midnight-zk/issues/278

Add proof-size tests (on all the examples) under a feature.


iquerejeta (commented):
Why do we add a feature for this? We should always be running the test - what is the issue with that? We already have a test for that, see [here](https://github.com/midnightntwrk/midnight-zk/blob/next/proofs/src/dev/cost_model.rs#L794). Clearly the test did not cover all edge-cases, but we could still do something similar, were we create a proof for a small circuit, with std_lib features, and check that the size matches.

PR #282, opened 2026-03-24
Adapt `foreign_params_gen.py` to the case where `a != 0`

John-Gong-Math:
To generate params for p256 base field over bls scalar field, run 
`python3 foreign_params_gen.py bls12-381-scalar secp256r1-base secp256r1`

To generate params for p256 scalar field over bls scalar field, run 
`python3 foreign_params_gen.py bls12-381-scalar secp256r1-scalar` (no need for curve info)

### DIFF
--- a/circuits/scripts/foreign_params_gen.py
+++ b/circuits/scripts/foreign_params_gen.py
@@ -155,7 +155,7 @@ def on_curve_expr_bounds(q, n, B, base_powers, double_base_powers, a_plus_one=1,
 
     # Handle the sign of (a+1) when computing min/max.
     # Note: (a+b) is a constant offset (not a coefficient of a variable), so it
-    # shifts both min and max equally regardless of sign — no case split needed.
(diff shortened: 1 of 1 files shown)


miguel-ambrona:
> Not sure if a or b become the leading factor, but the params of p256 generated by the initial script indeed failed before the changes.

I see. Do you remember what the params were? I would like to see the difference between those and the current ones.

circuits/scripts/foreign_params_gen.py
```diff
@@ -156,12 +156,8 @@ def on_curve_expr_bounds(q, n, B, base_powers, double_base_powers, a_plus_one=1,
     # Handle the sign of (a+1) when computing min/max.
     # Note: (a+b) is a constant offset (not a coefficient of a variable), so it
     # shifts both min and max equally regardless of sign — no case split needed.
```
miguel-ambrona: Remove "no case split needed" since there is no split now.

PR #288, opened 2026-03-24
Unify `nb_` and `nr_` as `num_`

miguel-ambrona:
Unify `nb_` and `nr_` as `num_` across all crates.

### DIFF
--- a/proofs/src/dev/cost_model.rs
+++ b/proofs/src/dev/cost_model.rs
@@ -266,15 +266,15 @@ impl CostOptions {
         // - COMM bytes for proof
         let multiopen = comp_bytes(2, point_sets);
 
-        let mut nr_rotations = HashSet::new();
+        let mut num_rotations = HashSet::new();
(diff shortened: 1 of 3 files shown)


davidnevadoc:
I'm on board with this change but I think we should make sure it will not create issues when we sync `main` and `next`.
If it creates too many conflicts, it may be a better option to wait until these branches are sync and then apply several renaming PRs like this one in one go.

miguel-ambrona:
> I'm on board with this change but I think we should make sure it will not create issues when we sync `main` and `next`. If it creates too many conflicts, it may be a better option to wait until these branches are sync and then apply several renaming PRs like this one in one go.

Yes, let's keep this on hold until `next` becomes more stable.

iquerejeta:
> Yes, let's keep this on hold until `next` becomes more stable.

Should we then close the PR for now?

miguel-ambrona:
Closing it for now.

PR #293, opened 2026-03-26
It is here - MockProver without k

iquerejeta:
We create a pass to the synthesize function that counts the number of rows needed for the circuit. 

We also tried making the mock prover have a dynamic structure, using vectors or hashmaps, and keep adding rows as we add constraints. However, that was uglier and less performant. This solution only sees ~25% regression in the mock prover.


proofs/src/dev/mod.rs
```diff
@@ -594,63 +585,249 @@ impl<F: Field> Assignment<F> for MockProver<F> {
...
+        &self,
+        column: Column<Instance>,
+        row: usize,
+    ) -> Result<circuit::Value<F>, Error> {
+        Ok(self
+            .instance
+            .get(column.index())
+            .and_then(|col| col.get(row))
+            .map(|v| circuit::Value::known(*v))
+            .unwrap_or_else(circuit::Value::unknown))
```
miguel-ambrona, line 721: I all this needed for inferring `k`?

proofs/src/dev/mod.rs
```diff
@@ -594,63 +585,249 @@ impl<F: Field> Assignment<F> for MockProver<F> {
...
+            circuit,
+            config.clone(),
+            constants.clone(),
+        )?;
+
+        let blinding_factors = cs.blinding_factors();
+        let required_n = (sizer.max_row + blinding_factors + 2).max(cs.minimum_rows());
+        let n = required_n.next_power_of_two();
+        let k = n.trailing_zeros();
+
```
miguel-ambrona, line 820: Instead of this block, can you just call the nice function `min_k` you defined above?

proofs/src/dev/mod.rs
```diff
@@ -712,6 +887,18 @@ impl<F: FromUniformBytes<64> + Ord> MockProver<F> {
         Ok(prover)
     }
 
+    /// Wrapper around [`MockProver::run_dynamic`] for API compatibility.
+    ///
+    /// The `k` parameter is ignored; the minimum required `k` is always determined
+    /// automatically from the circuit layout via [`RowSizer`].
+    pub fn run<ConcreteCircuit: Circuit<F>>(
+        _k: u32,
```
miguel-ambrona: Why don't we just remove argument `k`? This is not a trait function, is it?

proofs/benches/dev_lookup.rs
miguel-ambrona, line 1: Why did this go? Not used?

circuits/src/biguint/biguint_gadget.rs
```diff
@@ -1095,9 +1095,8 @@ mod tests {
             operation,
             _marker: PhantomData,
         };
-        let log2_nb_rows = 12;
         let public_inputs = vec![vec![], vec![]];
-        match MockProver::run(log2_nb_rows, &circuit, public_inputs) {
```
miguel-ambrona, line 1100: Why don't you keep the name "run"?

miguel-ambrona (approved):
This made my day!

I wonder if we can now infer the optimal k in zk_stdlib with this new dynamic_run instead of calling the (probably much slower) cost model. 😃

circuits/src/biguint/biguint_gadget.rs
miguel-ambrona: Sorry, pointing to the -, I was referring to the added code where you use `run_dynamic`.

proofs/benches/dev_lookup.rs
iquerejeta: We never benchmarked the lookup argument, and the benchmark was relying on fixing `k`.

circuits/src/biguint/biguint_gadget.rs
iquerejeta: Yes, I'm planning to keep the name `run`, but I'll need to handle the dependencies (blake and keccak), and it was a bit of a pain. I will make those changes before the merge.

proofs/src/dev/mod.rs
iquerejeta: Just matter of compatibility - blake and keccak would break otherwise. Once we agree on the final design of this, I'll update their dependencies  and replace this function.

proofs/src/dev/mod.rs
iquerejeta: good catch - instance values also account for nr of rows (or should I say `num  of rows`)

proofs/benches/dev_lookup.rs
miguel-ambrona: So now we don't care about that benchmark?

iquerejeta: No

PR #296, opened 2026-03-27
Create markers to be used as start and end point of circuits for the cost model

### LINKED ISSUE #289: Improve goldenfile tests to better reflect benchmarked functions
Currently, our goldenfile tests include assignment operations when measuring performance, which makes the results less representative of the actual function being benchmarked.

We should update the tests so they more accurately isolate and reflect the function’s execution, excluding unrelated operations like assignments where possible.

iquerejeta:
## Summary
- Add circuit benchmarking markers (start/end points) for the cost model
- Add new `field_chip_k256_base` entries to the cost-model golden file
- Update row counts across multiple circuit benchmarks to reflect the new marker boundaries

## Details
Introduces a mechanism to mark specific regions of circuits as start and end
points for benchmarking, enabling more accurate cost model measurements.
The cost-model golden file is updated with recalculated row counts and new
`field_chip_k256_base` circuit entries.

Closes #289


circuits/src/utils/circuit_modeling.rs
```diff
@@ -18,15 +18,45 @@ use std::{
...
 };
 use serde_json::{json, Map, Value};
 
+/// Marks the start of the cost-model measurement window inside a `synthesize`
+/// function.
+///
+/// Place this call immediately **before** the operation whose row cost you want
+/// to isolate. Pair with [`cost_measure_end`]. When both markers are
+/// present, the cost-model will report `rows` as the span covered by the
+/// marked section rather than the full circuit row count.
```
miguel-ambrona, line 37: What if only one is present or none?

iquerejeta: If just the beginning is set, the circuit will count till the end. If only the end is set, the full circuit will be counted (as the beginning was not set)

miguel-ambrona: Can you clarify that in the doc?

proofs/src/dev/cost_model.rs
```diff
@@ -314,26 +328,53 @@ pub(crate) fn cost_model_options<F: Ord + Field + FromUniformBytes<64>, C: Circu
...
+
+                    // A region is a _table region_ if all of its columns are `Fixed`
+                    // columns (see that [`plonk::circuit::TableColumn` is a wrapper
+                    // around `Column<Fixed>`]). All of a table region's rows are
+                    // counted towards `table_rows_count.`
+                    if region.columns.iter().all(|c| *c.column_type() == Fixed) {
+                        table_rows_count = std::cmp::max(table_rows_count, end + 1);
+                    } else {
+                        rows_count = std::cmp::max(rows_count, end + 1);
+                    }
```
miguel-ambrona: Can these two blocks be combined?

They are very similar.

circuits/src/hash/poseidon/poseidon_chip.rs
miguel-ambrona: This can now be removed.

miguel-ambrona (dismissed):
LGTM!

The cost-model.json seems to have new entries, did you add new tests? I missed that if you did. Can you try to remove the cost-model.json and run the tests again?
Also, some of the cost-model.json entries were renamed. Do you understand why?
EDIT: Ok, I see your comment in the PR description, I guess the diff is messed up.

This fine-grained measurement of the cost is very nice. I wonder what happens if a FloorPlanner is applied, does it mess up the count?

miguel-ambrona (approved):

PR #302, opened 2026-03-31
Add support for "resolving" the fixed bases in Accumulators and MSMs

### LINKED ISSUE #194: Enable an AssignedAccumulator to fully collapse
Add a `fully_collapse` method for `Accumulator`/`AssignedAccumulator` that evaluates both MSM sides (lhs/rhs) completely in-circuit by expanding the deferred `fixed_base_scalars` into explicit MSM terms using a provided `fixed_bases: BTreeMap<String, Point>` and then collapsing the result to two concrete points. Unlike the existing `collapse` (which only collapses the variable-base part), `fully_collapse` consumes the fixed-base part as well.

This is useful for “decider” circuits: keep the accumulator trick through recursive aggregation, then pay the fixed-base MSM cost once inside the final proof to output points suitable for a pairing-only check (rather than doing fixed-base work on-chain).

miguel-ambrona:
Fixes https://github.com/midnightntwrk/midnight-zk/issues/194


iquerejeta (approved):
LGTM 👍

PR #303, opened 2026-03-31
Fix `assigned_to_le` functions in field chip

remalue:
# The Problem
When calling `assigned_to_le_bytes` or `assigned_to_le_bits` with `nb_bits` exceeding `K::NUM_BITS`, the functions cause an index-out-of-bound error when trying to access the `bits` vector at an invalid index:

```
// Drop the most significant bits up to the desired length, but make sure they encode 0.
bits[nb_bits..]
    .iter()
    .try_for_each(|byte| self.native_gadget.assert_equal_to_fixed(layouter, byte, false))?;
bits = bits[0..nb_bits].to_vec();
```
# The Proposed Fix
Always cap the specified bit length via `nb_bits` at `K::NUM_BITS`:

```
let nb_bits = min(
    K::NUM_BITS as usize,
    nb_bits.unwrap_or(K::NUM_BITS as usize),
);
```

# Another Solution
If `nb_bits` is strictly greater than `K::NUM_BITS`, pad the resulting vector with 0 bits up to the requested length. But in the function `assigned_to_le_bits` panic if `enforce_canonical && nb_bits > K::NUM_BITS`.

### DIFF
--- a/circuits/src/field/foreign/field_chip.rs
+++ b/circuits/src/field/foreign/field_chip.rs
@@ -1158,7 +1158,7 @@ where
         bits = bits[0..nb_bits].to_vec();
 
         // The case nb_bits > K::NUM_BITS cannot happen since by above definition
-        // nb_bits = min(K::NUM_BITS,...), and thus nb_bits <= K::NUM_BITS
+        // nb_bits = min(K::NUM_BITS,...), and thus nb_bits <= K::NUM_BITS.
         if enforce_canonical && nb_bits == K::NUM_BITS as usize {
             let canonical = self.is_canonical(layouter, &bits)?;
             self.assert_equal_to_fixed(layouter, &canonical, true)?;
--- a/circuits/src/field/foreign/field_chip.rs
+++ b/circuits/src/field/foreign/field_chip.rs
@@ -1145,14 +1145,19 @@ where
             .into_iter()
             .for_each(|new_bits| bits.extend(new_bits));
 
+        let nb_bits = min(
+            K::NUM_BITS as usize,
+            nb_bits.unwrap_or(K::NUM_BITS as usize),
+        );
+
         // Drop the most significant bits up to the desired length, but make sure
         // they encode 0.
-        let nb_bits = nb_bits.unwrap_or(K::NUM_BITS as usize);
         bits[nb_bits..]
             .iter()
             .try_for_each(|byte| self.native_gadget.assert_equal_to_fixed(layouter, byte, false))?;
-        let bits = bits[0..nb_bits].to_vec();
-        if enforce_canonical && nb_bits >= K::NUM_BITS as usize {
+        bits = bits[0..nb_bits].to_vec();
+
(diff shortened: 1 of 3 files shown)


miguel-ambrona (approved):
LGTM.

circuits/src/field/foreign/field_chip.rs
```diff
@@ -1145,14 +1145,19 @@ where
...
         // they encode 0.
-        let nb_bits = nb_bits.unwrap_or(K::NUM_BITS as usize);
         bits[nb_bits..]
             .iter()
             .try_for_each(|byte| self.native_gadget.assert_equal_to_fixed(layouter, byte, false))?;
-        let bits = bits[0..nb_bits].to_vec();
-        if enforce_canonical && nb_bits >= K::NUM_BITS as usize {
+        bits = bits[0..nb_bits].to_vec();
+
+        if enforce_canonical && nb_bits == K::NUM_BITS as usize {
```
miguel-ambrona, line 1162: Either leave the `>=` or add a comment explaining that `>` cannot happen given the above `nb_bits = min(K::NUM_BITS, ...)`.

I wanna tie both code blocks (with such comment) so that they don't evolve independently, as one depends on the other.

remalue: I'll add a comment 👍

circuits/src/instructions/decomposition.rs
```diff
@@ -111,7 +112,8 @@ where
     /// constraints.
     ///
     /// The number of bytes (the length of the resulting vector) can be
-    /// specified. If unspecified, the resulting vector will contain exactly
+    /// specified, but will be capped at `ceil(Self::Assigned::Element::NUM_BITS
+    /// / 8)`. If unspecified, the resulting vector will contain exactly
```
miguel-ambrona: Do not split formulas in 2 lines please.

circuits/src/field/foreign/field_chip.rs
```diff
@@ -1145,14 +1145,21 @@ where
...
-        let nb_bits = nb_bits.unwrap_or(K::NUM_BITS as usize);
         bits[nb_bits..]
             .iter()
             .try_for_each(|byte| self.native_gadget.assert_equal_to_fixed(layouter, byte, false))?;
-        let bits = bits[0..nb_bits].to_vec();
-        if enforce_canonical && nb_bits >= K::NUM_BITS as usize {
+        bits = bits[0..nb_bits].to_vec();
+
+        // The case nb_bits > K::NUM_BITS cannot happen since by above definition
+        // nb_bits = min(K::NUM_BITS,...), and thus nb_bits <= K::NUM_BITS
```
miguel-ambrona:
```suggestion
        // The case nb_bits > K::NUM_BITS cannot happen since by above definition
        // nb_bits = min(K::NUM_BITS,...), and thus nb_bits <= K::NUM_BITS.
```

PR #305, opened 2026-04-01
Optimize Foreign Edwards MSM

davidnevadoc:
## Windowed MSM for foreign Edwards curves

This PR introduces windowed scalar multiplication for foreign Edwards ECC.

  - **Shared MSM preprocessing** (common.rs): Extracts the preprocessing logic (identity filtering, 1-bit scalar separation, base/scalar deduplication)  from `weierstrass_chip` into a shared generic function used by both Edwards and Weierstrass chips.
  - **Windowed Edwards MSM**: Implements `precompute`, `multi_select` , `windowed_msm` and  `windowed_scalar_mul` on the Edwards chip. 
   `msm()` uses shared-doubling with WS=3;
   `msm_by_bounded_scalars()` uses per-base windowed mul with adaptive   window size (WS=1/2/3 based on bit-bound).
  - Optimized Edwards doubling (by optimizing `add`): When p == q, reuses `p.x * q.y` for `p.y * q.x`, saving one foreign-field multiplication per doubling-as-add.
    
    ## TODO
    - [x] Drop note commit
    - [ ] Update changelog


circuits/goldenfiles/cost-model.json
```diff
@@ -584,7 +584,7 @@
       "max_deg": "5",
       "permutations": "32",
       "point_sets": "5",
-      "rows": "216566",
+      "rows": "87164",
```
remalue: 🚀

circuits/src/ecc/foreign/common.rs
```diff
@@ -0,0 +1,113 @@
...
+    instructions::{AssignmentInstructions, EccInstructions, ScalarFieldInstructions},
+    types::InnerValue,
+    CircuitField,
+};
+
+/// Preprocessing of MSM inputs.
+///
+/// Phase 1: Filters out identity bases and separates 1-bit scalars.
+/// Phase 2: Deduplicates bases and scalars known to be equal.
+pub(crate) fn msm_preprocess<F, C, EI, SFI>(
```
miguel-ambrona, line 46: Why not `EI` -> `E` and `SFI` -> `S`.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -875,6 +875,116 @@ where
...
+
+        // Precompute tables for each base.
+        let tables: Vec<PrecomputedTable<F, C, B, WS>> = bases
+            .iter()
+            .map(|b| self.precompute::<WS>(layouter, b))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut res = self.assign_fixed(layouter, C::CryptographicGroup::identity())?;
+        for w in (0..num_windows).rev() {
+            // Skip doubling in first window.
```
miguel-ambrona: I would say in the "least-significant" window, instead of "first". I understand what you mean by "first", but you are processing it the last!

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -1367,4 +1367,112 @@ mod tests {
             chip.load_from_scratch(&mut layouter)
         }
     }
+
+    /// Test circuit for `windowed_msm`: computes sum_i(scalar_i * base_i) and
+    /// asserts the result equals the expected point.
+    #[derive(Clone, Debug)]
+    struct WindowedMsmCircuit<C: EdwardsCurve> {
```
miguel-ambrona: Is this test necessary? We are gonna test the MSM through generic tests anyway.

circuits/src/instructions/ecc.rs
```diff
@@ -110,6 +110,15 @@ where
...
         // This blanket implementation simply ignores all bounds.
+
+        // NOTE FOR Reviewers:
+        // Should it be the other way around?
+        // msm() calls msm_by_bounded_scalars() with max NUM_BITS
+        // instead of this function calling msm()?
+        // Right now foreign edwards_chip and foreign weierstrass chips implement both functions,
+        // but if their msm_by_bounded_scalars
+        // impl is removed we would get a circular function call.
+
```
miguel-ambrona: I think it is ok as it currently is.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -738,50 +743,83 @@ where
...
+        // Add 1-bit scalar bases.
+        bases_with_1bit_scalar.iter().try_fold(res, |acc, (b, s)| {
+            let s_times_b = if s == &one {
+                b.clone()
+            } else {
+                let s_is_zero = scalar_chip.is_zero(layouter, s)?;
+                self.select(layouter, &s_is_zero, &identity, b)?
+            };
+            self.add(layouter, &acc, &s_times_b)
+        })
```
miguel-ambrona: Could this part also be added to the `common.rs` file and shared across chips?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -738,50 +743,83 @@ where
...
     }
 
-    // This function currently implements a basic form of double-and-add.
-    // There are several improvements available:
-    //  * Batching equal points
-    //  * Filtering scalars (e.g., if they are 0 or 1)
-    //  * Using the windowed method
+    // Per-base windowed MSM with adaptive window size. Unlike `msm`, this does
+    // not share doublings across bases, avoiding padding waste when scalar
+    // bounds differ significantly.
```
miguel-ambrona, line 734: Better than that, we can share the doubling across bases, but skip the internal addition of points whose scalar has been consumed. This is not done in the Weierstrass chip because it is not possible due to the random point `r` (for incomplete addition). But in Edwards, with complete addition, nothing stops you!

miguel-ambrona (commented):
Looks good!

I was expecting a dynamic lookup to load the table points given the scalar window, just what we do in the Weierstrass chip. In that case I would expect an optimal `WS = 4`.

This dynamic lookup + the idea I mentioned on sharing the doublings can lead to much better numbers. WDYT?

circuits/src/ecc/foreign/edwards_chip.rs
davidnevadoc: In this scenario, choosing a specific window size for each scalar ( depending on its size ) was more beneficial that sharing the doubling. Of course, both can be done at once, but that requires re-implementing or copy-pasting code from `windowed_scalar_mul` and its fairly complicated.

Which approach is the most optimal depends on the specific MSM characteristics. For example, right now the goldenfiles for this function are evaluated with an MSM that is, IMO, weird and unusual. It has only 3 scalars, with bit lengths 3, 12 and 255. It is therefore quite short and very decompensated and amortizing the doubling is not worth it.

I think it is fine to leave this as is and optimize if  we get a clearer use case target that performs badly with the 2 versions that are currently offered.

circuits/src/ecc/foreign/common.rs
davidnevadoc: `EI` for EccInstructions
`SFI` for ScalarFieldInstructions
I to indicate that these are generic instructions.
We can also go `NativeGadget` style and write the full thing in the generics: `EccArith` and `ScalarArith`, something like that.

davidnevadoc:
> Looks good!
> 
> I was expecting a dynamic lookup to load the table points given the scalar window, just what we do in the Weierstrass chip. In that case I would expect an optimal `WS = 4`.
> 
> This dynamic lookup + the idea I mentioned on sharing the doublings can lead to much better numbers. WDYT?

Oh, that was not my approach to the change at all!
I thought only of optimizations that reduced rows without changing the circuit architecture. The main reason being that, unlike the Weierstrass chip, which is critical for self-recursion. I think it is very unlikely that we amortize the cost of adding a lookup to save rows in a foreign Edwards MSM.

circuits/src/ecc/foreign/edwards_chip.rs
davidnevadoc: Removed!

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -857,6 +895,131 @@ where
...
+            for i in 0..half_len {
+                res[i] = self.select(layouter, b, &res[i + half_len], &res[i])?;
+            }
+            len = half_len;
+        }
+        Ok(res[0].clone())
+    }
+
+    /// Windowed interleaved MSM over LE bit-decomposed scalars. Shares the
+    /// doubling chain across all bases. MSB-first Horner evaluation.
```
remalue: Since you use iterative doubling, Horner evaluation is always "MSB-first".

Maybe just Horner evaluation?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -857,6 +895,131 @@ where
...
+        for _ in 2..1 << WS {
+            acc = self.add(layouter, &acc, base)?;
+            table.push(acc.clone());
+        }
+        Ok(PrecomputedTable { table })
+    }
+
+    /// Selects `table[w]` where `w` is the LE value of `bits`, using a binary
+    /// mux tree. Cost: 2^WS - 1 point selects.
+    fn multi_select<const WS: usize>(
```
remalue, line 883: That's nice! No adds necessary.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -875,6 +875,116 @@ where
...
+        bases: &[AssignedForeignEdwardsPoint<F, C, B>],
+    ) -> Result<AssignedForeignEdwardsPoint<F, C, B>, Error> {
+        assert_eq!(scalars.len(), bases.len());
+
+        if bases.is_empty() {
+            return self.assign_fixed(layouter, C::CryptographicGroup::identity());
+        }
+
+        // Pad all bit vectors to a common length that is a multiple of WS.
+        let max_bits = scalars.iter().map(|s| s.len()).max().unwrap_or(0);
```
remalue, line 918: What's the rationale for `unwrap_or(0)`?

circuits/src/ecc/foreign/common.rs
remalue: Is it possible to expand on the doc of this function? What does it return, and what are the elements in the return?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -738,50 +743,83 @@ where
...
+                b.clone()
+            } else {
+                let s_is_zero = scalar_chip.is_zero(layouter, s)?;
+                self.select(layouter, &s_is_zero, &identity, b)?
+            };
+            self.add(layouter, &acc, &s_times_b)
+        })
     }
 
     fn mul_by_constant(
```
remalue, line 770: Is there a reason why you didn't use the windowed method here?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -738,50 +743,83 @@ where
...
-            for (i, b) in scalar_bits.iter().enumerate() {
-                let addend = self.select(layouter, b, &p, &identity)?;
-                res = self.add(layouter, &res, &addend)?;
-                // The doubling in the last iteration is not needed
-                if i < scalar_bits.len() - 1 {
-                    p = self.double(layouter, &p)?;
-                }
-            }
+        // We do not share doublings because we assume the difference in
+        // scalar upper bounds negate the benefits, since we will double
```
remalue, line 751: Is this obvious in practice? Or was the thought to have the shared doublings in the `msm` and a non-shared version in here?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -738,50 +743,83 @@ where
...
+        let scalar_bits = scalars
             .iter()
-            .map(|s| (s.clone(), C::ScalarField::NUM_BITS as usize))
-            .collect::<Vec<_>>();
-
-        self.msm_by_bounded_scalars(layouter, &scalars, bases)
+            .map(|s| scalar_chip.assigned_to_le_bits(layouter, &s.0, None, true))
+            .collect::<Result<Vec<_>, Error>>()?;
+
+        // Windowed msm with shared doubling.
```
remalue, line 725: Could you move this comment above the function itself? This way it's immediately visible.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: Done

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: I'll address this in a subsequent PR.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: I guess not panicking. Although an unwrap would do too, since `scalars` is not empty.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: In `mul_by_constant` it is usually better to do things with 1-bit windows. That way on average, you skip half of the additions.

If we used e.g. `WS = 3` you would only skip 1 out of 8 additions (when the window is 000).
On the other hand you are doing roughly 1/3 of the additions. I'll play with different options in a subsequent PR.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: I'll share the doublings in a subsequent PR.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: What function? The `msm`? That comment does not belong there IMHO. Also, we tend not to document trait functions in their implementation.

circuits/src/ecc/foreign/common.rs
miguel-ambrona: Done.

circuits/src/ecc/foreign/edwards_chip.rs
remalue: I am not speaking about doc. The function `msm_by_bounded_scalars` has a comment as well, and I think it's good to have.

circuits/src/ecc/foreign/edwards_chip.rs
remalue: Yes, just unwrap since it's not empty.

remalue (approved):
Excellent optimization! 🚀

Left some minor comments.

PR #306, opened 2026-04-01
Add shared column pools to `configure_from_scratch`

### LINKED ISSUE #152: Reusing the assign chip for the test circuit
For `SHA256` test and cost model [here](https://github.com/midnightntwrk/midnight-zk/blob/main/circuits/src/hash/sha256/mod.rs#L106-L114), `native-gadget` is configured twice, one for `SHA256` and one for assign chip of the test circuit. We may reuse the `native-gadget` defined within `sha256` chip as the assign chip for the test circuit.

miguel-ambrona:
Fixes https://github.com/midnightntwrk/midnight-zk/issues/152


circuits/src/biguint/biguint_gadget.rs
```diff
@@ -19,6 +19,8 @@ use std::{
     ops::Rem,
 };
 
+#[cfg(any(test, feature = "testing"))]
+use midnight_proofs::plonk::{Advice, Fixed};
```
iquerejeta: You already have a block with these conditional compilation flag below. Move this there (line31).

circuits/src/instructions/equality.rs
```diff
@@ -154,6 +154,8 @@ pub(crate) mod tests {
             meta.enable_constant(constants_column);
             EqualityChip::configure_from_scratch(
                 meta,
+                &mut vec![],
+                &mut vec![constants_column],
```
iquerejeta: can we share the constants column? We don't do that in the stdlib

iquerejeta (dismissed):
Great work. Now the goldenfiles are more realistic! Just minor comments.

circuits/src/instructions/equality.rs
miguel-ambrona: In this chip is called "constants" but it is used as a normal fixed column IIRC.

But good point, I wonder whether we could use a normal shared fixed column for the constants column in zk_stdlib.

miguel-ambrona: Ok had a look and I think it is a bad idea to mix columns for constants with normal fixed columns.

I'm reverting this in the last commit. Indeed, these tests (unlike the `control_flow` tests) use "constrain_constant" and not "assign_fixed".

iquerejeta (approved):

PR #307, opened 2026-04-01
Deterministic prover

iquerejeta:
Minor version bump with deterministic prover and new function to extract g2 from verifier params.

### DIFF
--- a/circuits/src/ecc/foreign/ecc_chip.rs
+++ b/circuits/src/ecc/foreign/ecc_chip.rs
@@ -950,8 +950,8 @@ where
 {
     /// Given config creates new chip that implements foreign ECC.
     /// The RNG is used to sample a random point used for incomplete addition
-    /// during windowed MSM operations. Randomness does not affect soundness,
-    /// only completeness.
+    /// Soundness does not rely on this point being random, but completeness 
+    /// does.
     pub fn new(
         config: &ForeignEccConfig<C>,
         native_gadget: &N,
--- a/circuits/src/ecc/foreign/ecc_chip.rs
+++ b/circuits/src/ecc/foreign/ecc_chip.rs
@@ -950,7 +950,7 @@ where
 {
     /// Given config creates new chip that implements foreign ECC.
     /// The RNG is used to sample a random point used for incomplete addition
-    /// Soundness does not rely on this point being random, but completeness 
+    /// Soundness does not rely on this point being random, but completeness
     /// does.
     pub fn new(
         config: &ForeignEccConfig<C>,
--- a/circuits/src/ecc/foreign/ecc_chip.rs
+++ b/circuits/src/ecc/foreign/ecc_chip.rs
@@ -24,7 +24,6 @@ use std::{
     collections::HashMap,
     fmt::Debug,
     hash::{Hash, Hasher},
-    ops::Mul,
     rc::Rc,
 };
 
@@ -120,8 +119,9 @@ where
     // It will never overflow unless you include more than 2^64 tables, will you?
     // Even in that case, we would get a compile-time error.
     tag_cnt: Rc<RefCell<u64>>,
-    // A random point used in windowed_msm to allow incomplete addition.
-    // Sampled once at chip construction time.
+    // A random point used in windowed_msm (to shift the initial accumulator) so that the
+    // double-and-add loop can internally use incomplete addition. Sampled once at chip
+    // construction time.
     random_point: C::CryptographicGroup,
 }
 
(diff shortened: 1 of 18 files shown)


proofs/src/poly/kzg/params.rs
```diff
@@ -242,6 +242,11 @@ impl<E: MultiMillerLoop + Debug> ParamsVerifierKZG<E>
 where
     E::G2: Curve + ProcessedSerdeObject,
 {
+    /// Returns first power of secret on G2
```
miguel-ambrona: This sounds weird.

proofs/src/plonk/prover.rs
```diff
@@ -368,14 +369,15 @@ where
         + Ord
         + FromUniformBytes<64>,
 {
+    let mut rng = rng;
```
miguel-ambrona, line 371: Hum, why not letting `create_proof` get the `rng` mutable?

circuits/src/ecc/foreign/ecc_chip.rs
```diff
@@ -120,6 +120,9 @@ where
     // It will never overflow unless you include more than 2^64 tables, will you?
     // Even in that case, we would get a compile-time error.
     tag_cnt: Rc<RefCell<u64>>,
+    // A random point used in windowed_msm to allow incomplete addition.
```
miguel-ambrona: "allow" is not the best word.

I would say:
"A random point used windowed_msm (to shift the initial accumulator) so that the double-and-add loop can internally use incomplete addition without affecting the overall completeness".

Maybe you can drop "without affecting the overall completeness".

circuits/src/ecc/foreign/ecc_chip.rs
```diff
@@ -945,15 +948,26 @@ where
     S::Scalar: InnerValue<Element = C::Scalar>,
     N: NativeInstructions<F>,
 {
-    /// Given config creates new chip that implements foreign ECC
-    pub fn new(config: &ForeignEccConfig<C>, native_gadget: &N, scalar_field_chip: &S) -> Self {
+    /// Given config creates new chip that implements foreign ECC.
+    /// The RNG is used to sample a random point used for incomplete addition during
+    /// windowed MSM operations.
```
miguel-ambrona: Explain that the point does not need to be random for the sake of soundness, it is only for completeness.

circuits/src/ecc/foreign/ecc_chip.rs
```diff
@@ -945,15 +948,26 @@ where
...
+    pub fn new(
+        config: &ForeignEccConfig<C>,
+        native_gadget: &N,
+        scalar_field_chip: &S,
+        mut rng: impl RngCore + CryptoRng,
+    ) -> Self {
+        let r_dlog = C::Scalar::random(&mut rng);
+        let random_point =
+            C::CryptographicGroup::mul(C::CryptographicGroup::generator(), r_dlog);
```
miguel-ambrona: Use the occasion to sample a point directly with `C::random`. I already did this in `main` and we may lose those changes if we are not careful when rebasing this.

miguel-ambrona (approved):
Good. It was not that bad in the end.

What are we gonna do with these changes? Will the come to `main` eventually?

proofs/src/plonk/prover.rs
iquerejeta: That changes the interface, and requires a new major version. That'd be annoying.

circuits/src/ecc/foreign/ecc_chip.rs
```diff
@@ -950,15 +950,15 @@ where
 {
     /// Given config creates new chip that implements foreign ECC.
     /// The RNG is used to sample a random point used for incomplete addition
-    /// during windowed MSM operations.
+    /// during windowed MSM operations. Randomness does not affect soundness,
+    /// only completeness.
```
miguel-ambrona: Randomness does not affect soundness -> Randomness is not required for soundness.

Or even more explicit than that if you can:
Soundness does not rely on this point being random, but completeness does.

proofs/src/poly/kzg/params.rs
```diff
@@ -127,7 +127,7 @@ impl<E: Engine + Debug> ParamsKZG<E> {
         self.g2
     }
 
-    /// Returns first power of secret on G2
+    /// Returns [τ]₂, a commitment to τ in G2.
```
miguel-ambrona: "Commitment" sounds weird too, but it is understandable.

circuits/src/ecc/foreign/ecc_chip.rs
iquerejeta:
```suggestion
    /// Soundness does not rely on this point being random, but completeness 
    /// does.
```

proofs/src/poly/kzg/params.rs
iquerejeta:
```suggestion
    /// Returns \[τ\]₂, a commitment to τ in G2.
```

proofs/src/poly/kzg/params.rs
```diff
@@ -242,7 +242,7 @@ impl<E: MultiMillerLoop + Debug> ParamsVerifierKZG<E>
 where
     E::G2: Curve + ProcessedSerdeObject,
 {
-    /// Returns first power of secret on G2
+    /// Returns [τ]₂, a commitment to τ in G2.
```
iquerejeta:
```suggestion
    /// Returns \[τ\]₂, a commitment to τ in G2.
```

PR #286, opened 2026-03-24
Document end-of-synthesize table loading requirement (Issue 149)

kanyuku:
Added warnings to Circuit::synthesize, NativeGadget, and relevant chips (Pow2RangeChip, ScannerChip,KeccakSha3Wrapper) to ensure lookup tables are loaded at the very end of synthesis, preventing constraint system errors.

PR #309, opened 2026-04-02
fix(circuits): enforce table-load-after-emit invariant in Pow2RangeChip and ScannerChip

### LINKED ISSUE #149: Loading tables in a `Circuit` implementation has to be done at the end
### Description of Issue:

When implementing the `Circuit` trait, typically when writing tests for libraries using `FromScratch` without access to `ZkStdLib`, constraint system errors may arise if the tables are loaded anywhere else than the very end (of the `synthesize` method). The errors I got on my example suggests this only applied to `native_gadget`, but there is a possibility it also impacts other chips (since their table are only loaded if a reference has been updated upon witnessing their effective usage).

The only trace of a documentation of this behaviour I could find is [here](https://github.com/midnightntwrk/midnight-zk/blob/802389dfad3a5834a32ceece42e9eb6ddca7a98e/circuits/src/compact_std_lib.rs#L1536-L1537), but this is just a comment during the construction of Compact std lib.

### Description of Improvement:

It would be probably better to have this documented properly, maybe in the documentation of the `Circuit` trait?

doomhammerhell:
Closes #149

Problem
-------
Lookup tables in Pow2RangeChip and ScannerChip are dynamically built from state accumulated during constraint emission (queried_tags and automaton_cache, respectively). If load_table()/load() is called before all constraints are emitted, the materialized table is incomplete and lookup verification fails silently -- producing cryptic constraint system errors with no indication of the actual root cause.

This ordering dependency was implicit, undocumented, and enforced only by careful placement of load() calls at the end of synthesize(). Any new chip composer or FromScratch user could violate it unknowingly.

Root Cause
----------
- Pow2RangeChip::load_table() reads queried_tags (Rc/RefCell/HashSet) to decide which (tag, value) pairs to include in the lookup table. If called before constraint emission, queried_tags is empty and the table contains only the sentinel (tag=0, val=0).
- ScannerChip::load() reads automaton_cache and sequence_cache to build the automaton transition table and finalize substring checks. Same issue.

Solution
--------
Introduce a frozen: Rc/Cell/bool flag in both chips that converts the implicit temporal contract into an explicit, runtime-enforced invariant:

- load_table()/load() sets frozen=true before materializing the table.
- assert_row_lower_than_2_pow_n(), parse(), and check_bytes() panic with a clear, actionable error message if called after the chip is frozen.
- Added Pow2RangeChip::precompute_tags() as an escape hatch for circuits where the full tag set is known upfront.
- Added Pow2RangeChip::is_frozen() for state introspection.
- Updated ComposableChip::load() doc-comment with the explicit contract.

Files Changed
-------------
- circuits/src/field/decomposition/pow2range.rs
- circuits/src/parsing/scanner/mod.rs
- circuits/src/parsing/scanner/automaton_chip.rs
- circuits/src/parsing/scanner/substring.rs
- circuits/src/utils/composable.rs

Tests
-----
Added 4 new tests in pow2range.rs:
- test_post_load_constraint_panics (negative: panic on post-load emit)
- test_precompute_tags_enables_early_load (positive: precompute escape hatch)
- test_precompute_after_freeze_circuit (negative: panic on late precompute)
- test_is_frozen_state (state transition verification)

All 210 circuit unit tests, 4 zk_stdlib tests, and 49 doc tests pass.

Audit
-----
Full analysis in audit/issue-149-table-order-dependency/: ROOT_CAUSE.md, STATE_MACHINE.md, INVARIANTS.md, DESIGN_OPTIONS.md, FINAL_DESIGN.md, PATCH.md, TEST_PLAN.md

### DIFF
--- a/.git-commit-msg.txt
+++ b/.git-commit-msg.txt
@@ -1,0 +1,62 @@
+fix(circuits): enforce table-load-after-emit invariant in Pow2RangeChip and ScannerChip
+
+Closes #149
+
+Problem
+-------
+Lookup tables in Pow2RangeChip and ScannerChip are dynamically built from
+state accumulated during constraint emission (queried_tags and automaton_cache,
+respectively). If load_table()/load() is called before all constraints are
+emitted, the materialized table is incomplete and lookup verification fails
+silently -- producing cryptic constraint system errors with no indication of
+the actual root cause.
+
+This ordering dependency was implicit, undocumented, and enforced only by
+careful placement of load() calls at the end of synthesize(). Any new chip
+composer or FromScratch user could violate it unknowingly.
+
+Root Cause
+----------
+- Pow2RangeChip::load_table() reads queried_tags (Rc/RefCell/HashSet) to
+  decide which (tag, value) pairs to include in the lookup table. If called
+  before constraint emission, queried_tags is empty and the table contains
+  only the sentinel (tag=0, val=0).
+- ScannerChip::load() reads automaton_cache and sequence_cache to build the
+  automaton transition table and finalize substring checks. Same issue.
+
+Solution
+--------
+Introduce a frozen: Rc/Cell/bool flag in both chips that converts the
+implicit temporal contract into an explicit, runtime-enforced invariant:
+
+- load_table()/load() sets frozen=true before materializing the table.
+- assert_row_lower_than_2_pow_n(), parse(), and check_bytes() panic with a
+  clear, actionable error message if called after the chip is frozen.
+- Added Pow2RangeChip::precompute_tags() as an escape hatch for circuits
+  where the full tag set is known upfront.
+- Added Pow2RangeChip::is_frozen() for state introspection.
+- Updated ComposableChip::load() doc-comment with the explicit contract.
+
+Files Changed
+-------------
+- circuits/src/field/decomposition/pow2range.rs
+- circuits/src/parsing/scanner/mod.rs
+- circuits/src/parsing/scanner/automaton_chip.rs
+- circuits/src/parsing/scanner/substring.rs
+- circuits/src/utils/composable.rs
+
+Tests
+-----
+Added 4 new tests in pow2range.rs:
+- test_post_load_constraint_panics (negative: panic on post-load emit)
+- test_precompute_tags_enables_early_load (positive: precompute escape hatch)
+- test_precompute_after_freeze_circuit (negative: panic on late precompute)
+- test_is_frozen_state (state transition verification)
+
+All 210 circuit unit tests, 4 zk_stdlib tests, and 49 doc tests pass.
+
+Audit
+-----
+Full analysis in audit/issue-149-table-order-dependency/:
+ROOT_CAUSE.md, STATE_MACHINE.md, INVARIANTS.md, DESIGN_OPTIONS.md,
+FINAL_DESIGN.md, PATCH.md, TEST_PLAN.md
--- a/circuits/src/field/decomposition/pow2range.rs
+++ b/circuits/src/field/decomposition/pow2range.rs
@@ -14,7 +14,13 @@
 //! `pow2range` is a chip for performing membership assertions in ranges of the
 //! form [0, 2^n) via lookups.
 
-use std::{cell::RefCell, collections::HashSet, fmt::Debug, marker::PhantomData, rc::Rc};
+use std::{
+    cell::{Cell, RefCell},
+    collections::HashSet,
+    fmt::Debug,
+    marker::PhantomData,
+    rc::Rc,
+};
 
 use midnight_proofs::{
     circuit::{Chip, Layouter, Region, Value},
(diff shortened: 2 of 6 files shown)


circuits/src/field/decomposition/pow2range.rs
```diff
@@ -68,13 +74,27 @@ pub struct Pow2RangeConfig {
...
+/// [`precompute_tags`](Self::precompute_tags) to register them before
+/// `load_table`, which then makes the loading order safe.
+#[derive(Clone, Debug)]
 pub struct Pow2RangeChip<F: CircuitField> {
     config: Pow2RangeConfig,
     max_bit_len: usize,
     queried_tags: Rc<RefCell<HashSet<usize>>>,
+    /// Set to `true` once `load_table` has been called. Any subsequent call to
+    /// `assert_row_lower_than_2_pow_n` will panic.
+    frozen: Rc<Cell<bool>>,
```
miguel-ambrona, line 97: Would `loaded` be a better name instead of `frozen`?

circuits/src/field/decomposition/pow2range.rs
```diff
@@ -234,6 +273,40 @@ impl<F: CircuitField> Pow2RangeChip<F> {
...
+        for &tag in tags {
+            assert!(
+                tag <= self.max_bit_len,
+                "precompute_tags: tag={} exceeds max_bit_len={}",
+                tag,
+                self.max_bit_len
+            );
+            queried.insert(tag);
+        }
+    }
```
miguel-ambrona, line 303: Is this really necessary? If not, remove it.

.git-commit-msg.txt
miguel-ambrona, line 1: Do not commit this file please.

circuits/src/field/decomposition/pow2range.rs
```diff
@@ -367,4 +440,199 @@ mod tests {
...
+            _marker: PhantomData<F>,
+        }
+
+        impl<F: CircuitField> Circuit<F> for PostLoadCircuit<F> {
+            type Config = Pow2RangeConfig;
+            type FloorPlanner = SimpleFloorPlanner;
+            type Params = ();
+
+            fn without_witnesses(&self) -> Self {
+                unreachable!();
```
miguel-ambrona, line 461: These tests are too verbose.

Maybe you can leverage the existing tests and load the table at the beginning or at the end conditionally on a flag. WDYT?

miguel-ambrona (changes requested):
Thanks a lot @doomhammerhell .

This is a nice way to solve the problem.
Please, make sure you sign all your commits.

circuits/src/field/decomposition/pow2range.rs
```diff
@@ -102,6 +122,15 @@ impl<F: CircuitField> Pow2RangeChip<F> {
...
         }
+        if self.frozen.get() {
+            panic!(
+                "Pow2RangeChip: cannot emit lookup constraint for tag={} after load_table() \
+                 has been called. The lookup table has already been materialized and does not \
+                 contain this tag. Either move load_table() to after all constraint emission, \
+                 or use precompute_tags() to register all required tags before loading.",
+                n
+            )
+        }
```
irakoton, line 133: It would be cleaner to use an `assert!` (also in the block above actually, even if you are not the one who introduced it). I also think it would be much better to have a less explicit but shorter error message, and to remove the `precompute_tags` option. Applying `load` at the end is never an issue, so having a convoluted other option just harms the codebase readability. Something like "<chip name>: cannot emit lookup constraint after <load function> has been called. <load function> has to be called at the end of a circuit syntehsis."

circuits/src/field/decomposition/pow2range.rs
irakoton: I also think it's not useful.

circuits/src/parsing/scanner/automaton_chip.rs
```diff
@@ -408,12 +408,23 @@ where
...
         layouter: &mut impl Layouter<F>,
         parser: AutomatonParser,
         input: &[AssignedByte<F>],
     ) -> Result<Vec<AssignedNative<F>>, Error> {
+        if self.frozen.get() {
+            panic!(
+                "ScannerChip: cannot call parse() after load() has been called. \
+                 The automaton transition table has already been materialized. \
+                 Move load() to after all parse() calls."
+            )
```
irakoton, line 426: Same remark as for the decomposition chip: use an `assert!`, and a shorter error message.

circuits/src/parsing/scanner/substring.rs
```diff
@@ -207,13 +207,24 @@ where
...
         idx: &AssignedNative<F>,
         sub: &[AssignedByte<F>],
     ) -> Result<(), Error> {
+        if self.frozen.get() {
+            panic!(
+                "ScannerChip: cannot call check_bytes() after load() has been called. \
+                 The substring check table has already been finalized. \
+                 Move load() to after all check_bytes() calls."
+            )
+        }
```
irakoton, line 227: Same remark as for the decomposition chip: use an `assert!`, and a shorter error message.

Also, this is not the function that should have the frozen/loaded check. `check_subsequence` is the base function that the chip is relying on, and which should have this check. Future extensions of this substring checks will always call `check_subsequence` in some kind of context, similarly to `check_bytes` that is just a wrapper.

irakoton (changes requested):
I'm leaving here some comments if you want to improve the PR, and I also agree with @miguel-ambrona 's comments. It seems your commits are also not signed, which will block potential merge.

That being said, we will have to close this issue for the time being. There is already an anterior submission for the exact same problem [#286](https://github.com/midnightntwrk/midnight-zk/pull/286). Your solution is more complete, so we may still consider it in the future, if you address all the remarks.

In the future, keep however in mind that the normal workflow is to comment on the existing PR instead of creating a duplicate submission like this one.

irakoton:
Closing for the time being due to being a duplicate with #286. May be re-opened in the future if we find the additional value is worth it and comments are addressed.

PR #310, opened 2026-04-02
Ed25519 in-circuit signature verification

remalue:
Adds an example with `ed25519` in-circuit signature verification to `zk_stdlib` (Cardano uses `ed25519` signatures for, e.g., transaction signing).

# ⚠️ Checklist
- [x] Clarify if canonicity checks for $A$, $R$, and $s$ are needed.


iquerejeta:
> Clarify if canonicity checks for `A`, `R`, and `s` are needed.

We should use the same verification criteria as Cardano uses.

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+//!   * B is the designated generator of G1 (in C).
+//!   * A is the public key (in C).
+//!   * σ = (R,s) is the signature, with:
+//!     - R the nonce commitment (in C),
+//!     - s the signature scalar (in F_L).
+//!   * h = SHA-512(R_bytes || A_bytes || M_bytes) mod L is the challenge (in
+//!     F_L), with:
+//!     - R_bytes are the bytes of the compressed R,
+//!     - A_bytes are the bytes of the compressed A.
+//!     - M_bytes are the bytes for message M
```
miguel-ambrona: Specify endianess?

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn format_instance((public_key, msg): &Self::Instance) -> Result<Vec<F>, Error> {
+        Ok([
+            AssignedForeignEdwardsPoint::<F, Curve25519, MultiEmulationParams>::as_public_input(
+                public_key,
+            ),
+            msg.iter().flat_map(AssignedByte::<F>::as_public_input).collect::<Vec<_>>(),
+        ]
+        .into_iter()
+        .flatten()
+        .collect())
```
miguel-ambrona: With "concat", it looks a bit cleaner:

```rust
Ok([
    AssignedForeignEdwardsPoint::<F, Curve25519, MEP>::as_public_input(public_key),
    msg.iter().flat_map(AssignedByte::<F>::as_public_input).collect::<Vec<_>>(),
]
.concat())
```

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+use midnight_proofs::{
+    circuit::{Layouter, Value},
+    plonk::Error,
+};
+use midnight_zk_stdlib::{utils::plonk_api::filecoin_srs, Relation, ZkStdLib, ZkStdLibArch};
+use rand::rngs::OsRng;
+
+type F = midnight_curves::Fq;
+
+const MSG_LEN: usize = 86;
```
miguel-ambrona, line 66: Is this standard or an arbitrary choice?

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+        // Assign the message bytes M and constrain them as public input.
+        let m_bytes: Vec<AssignedByte<F>> =
+            std_lib.assign_many(layouter, &instance.map(|(_, msg)| msg).transpose_array())?;
+        m_bytes
+            .iter()
+            .try_for_each(|byte| std_lib.constrain_as_public_input(layouter, byte))?;
+
+        // Decompose witness w = (R, s) and assign (R, s) as curve point and scalar.
+        let r_val = witness.map(|(r, _)| r);
+        let s_val = witness.map(|(_, s)| s);
```
miguel-ambrona: Why not inlining these two as you did for the instance?

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+        let r = curve25519_curve.assign(layouter, r_val)?;
+        let s = curve25519_scalar.assign(layouter, s_val)?;
+
+        // Canonicity check for s.
+        let s_bits = curve25519_scalar.assigned_to_le_bits(layouter, &s, None, false)?;
+        let canonical = curve25519_scalar.is_canonical(layouter, &s_bits)?;
+        curve25519_scalar.assert_equal_to_fixed(layouter, &canonical, true)?;
```
miguel-ambrona:
```suggestion
        let s_bits = curve25519_scalar.assigned_to_le_bits(layouter, &s, None, true)?;
```

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    // The result is guaranteed to lie in the prime-order subgroup.
+    Curve25519Subgroup::from_edwards(edwards).expect("curve25519 subgroup point")
+}
+
+// In-circuit compression of `AssignedForeignEdwardsPoint` into
+// `[AssignedByte<F>; 32]`.
+//
+// Note: Only points with canonical coordinates satisfy the underlying
+// constraints.
```
miguel-ambrona: It is unclear what this means. These low-level comments may not be appropriate in an example.

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+        let s = curve25519_scalar.assign(layouter, s_val)?;
+
+        // Canonicity check for s.
+        let s_bits = curve25519_scalar.assigned_to_le_bits(layouter, &s, None, false)?;
+        let canonical = curve25519_scalar.is_canonical(layouter, &s_bits)?;
+        curve25519_scalar.assert_equal_to_fixed(layouter, &canonical, true)?;
+
+        // Compress R and A.
+        let r_compressed_bytes = compress_edwards_point(std_lib, layouter, &r)?;
+        let a_compressed_bytes = compress_edwards_point(std_lib, layouter, &a)?;
```
miguel-ambrona: "compress_edwards_point" -> "to_compressed_bytes"

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+        Some(255),
+        true,
+    )?;
+
+    // 3. Encoding the sign bit of x-coordinate (= x mod 2, i.e., the least
+    //    significant bit) into the most significant byte of the y-coordinate:
+    //    compressed_point[31] = y_bytes[31] + x_sign_bit * 128.
+    //
+    // (This is safe: y <= p = 2^255 - 19 - 1, which means y_bytes[31] <= 127;
+    // hence, adding 128 to the most significant byte causes _no_ overflow.)
```
miguel-ambrona: I would add these details in the documentation of this function and not inlined during its implementation. 

If you are following a specification, the details of why this is "safe" may not even be necessary, although it is nice to have them.

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        // Compute lhs = s * B.
+        let lhs = curve25519_curve.msm(layouter, &[s], &[b])?;
+
+        // Compute rhs = R + h * A.
+        let h_a = curve25519_curve.msm(layouter, &[h], &[a])?;
+        let rhs = curve25519_curve.add(layouter, &r, &h_a)?;
+
+        // Assert s * B = R + h * A.
+        curve25519_curve.assert_equal(layouter, &lhs, &rhs)
```
miguel-ambrona: Why not computing `h * A - s * B` together in an msm?

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,281 @@
...
+    // The result is guaranteed to lie in the prime-order subgroup.
+    Curve25519Subgroup::from_edwards(edwards).expect("curve25519 subgroup point")
+}
+
+// In-circuit compression of `AssignedForeignEdwardsPoint` into
+// `[AssignedByte<F>; 32]`.
+//
+// Note: Only points with canonical coordinates satisfy the underlying
+// constraints.
+fn compress_edwards_point(
```
miguel-ambrona: Maybe this function should be implemented in the Edwards chip. What do you think? We can implement `DecompositionInstrucitons` for points in that chip (at least when the generic is instantiated with Curve25519).

miguel-ambrona (dismissed):
LGTM!

zk_stdlib/examples/cardano_signature.rs
remalue: Just arbitrary for this example.

zk_stdlib/examples/cardano_signature.rs
remalue: I think it's helpful to have (even if it's a spec).

zk_stdlib/examples/cardano_signature.rs
remalue: I was considering that, but was doubting the soundness.

We are not using `s_bits` afterwards, only `s`. But when calling it this way, the underlying constraints still ensure `s` is canonical? And just call with `let _ = ... `?

zk_stdlib/examples/cardano_signature.rs
remalue: I used `s * B - h * A`.

zk_stdlib/examples/cardano_signature.rs
remalue: Good point. The major use case of EdDSA verification speaks in favor of it. For me it's unclear how, overall, the chip is used and if strongly necessary.

Personally, I like to have such functions as part of the lib.

zk_stdlib/examples/cardano_signature.rs
miguel-ambrona: The term "canonical" does not apply to `s`. What are you trying to avoid here?

zk_stdlib/examples/cardano_signature.rs
miguel-ambrona: Move it to the edwards_chip.

remalue: #327

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -19,9 +19,9 @@
 //!     - s the signature scalar (in F_L).
 //!   * h = SHA-512(R_bytes || A_bytes || M_bytes) mod L is the challenge (in
 //!     F_L), with:
-//!     - R_bytes are the bytes of the compressed R,
-//!     - A_bytes are the bytes of the compressed A.
-//!     - M_bytes are the bytes for message M
+//!     - R_bytes are the LE bytes of the compressed R,
+//!     - A_bytes are the LE bytes of the compressed A,
+//!     - M_bytes are the LE bytes for message M.
```
miguel-ambrona: Do the message bytes have an endianess? Should them be treated as a black-box instead?

remalue: It depends on how you define the message.

zk_stdlib/examples/cardano_signature.rs
remalue: Only integer values `0 <= s < q` should be accepted, where `q` is the foreign scalar field modulus of `edwards25519`.

miguel-ambrona: But here `s` is a witness and, if I am not mistaken, the only thing you do with it is to multiply another point by it. Therefore, those checks are meaningless.

remalue: Why? Otherwise the prover potentially can supply two different point encodings.

miguel-ambrona: Maybe the cleanest way to put it is that a `PoK` of a non-canonical `s` satisfying the checks implies the `PoK` of a canonical `s` satisfying the checks. Thus, that kind of canonicity is not necessary here.

This is actually clear from the types. `AssignedField<F, curve25519Scalar, _>` does not have canonicity checks (and I'm not sure how to define those for this type, which is representation-agnostic). Only if we represented the scalar as bits or bytes would those checks come into play.

remalue: > Maybe the cleanest way to put it is that a PoK of a non-canonical s satisfying the checks implies the PoK of a canonical s satisfying the checks. Thus, that kind of canonicity is not necessary here.

Yes, but the point is that non-canonical `s` should not satisfy the checks.

miguel-ambrona: Let's discuss it on a call since we do not seem to find an agreement here.

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -63,8 +65,8 @@ type F = midnight_curves::Fq;
 
 const MSG_LEN: usize = 86;
 type Message = [u8; MSG_LEN];
-type PublicKey = Curve25519Subgroup;
-type Signature = (Curve25519Subgroup, Scalar);
+type PublicKey = ([u8; 32], Curve25519Subgroup); // (compressed bytes of A, A)
```
miguel-ambrona, line 68: Provide just the bytes:
```
type PublicKey = [u8; 32]; // compressed bytes of A
```

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -94,28 +97,45 @@ impl Relation for CardanoSigExample {
...
+        // element s.
+        let s_bytes: Vec<AssignedByte<F>> =
+            std_lib.assign_many(layouter, &witness.map(|(_, s)| s).transpose_array())?;
+        let s_bits: Vec<AssignedBit<F>> = s_bytes
+            .iter()
+            .map(|byte| std_lib.assigned_to_le_bits(layouter, &byte.clone().into(), Some(8), false))
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
+            .flatten()
+            .collect();
```
miguel-ambrona, line 128: This looks tedious. Maybe it deserves a helper function somewhere in utils.

We can also implement conversion from `Vec<AssignedBytes>` to `Vec<AssignedBits>`.

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -154,21 +174,102 @@ impl Relation for CardanoSigExample {
     }
 }
 
-// Off-circuit decompression of `CompressedEdwardsY` into
-// `Curve25519Subgroup` point.
-fn decompress_point(bytes: &[u8; 32]) -> Curve25519Subgroup {
-    let compressed = CompressedEdwardsY(*bytes);
-    let edwards = compressed.decompress().expect("y coordinate of curve25519 point");
+// TODO: document this function.
+fn from_compressed_bytes(
```
miguel-ambrona, line 178: This function resulted very long and complex.

It feels that "to_compress_bytes" was simpler. If so, you can simply implement `from_compressed_bytes` by witnessing the result, converting it to bytes and asserting equality with the given bytes.

remalue: Good point, I'll think about it.

remalue: The approach that witnesses an assigned coordinate and then calls `to_assigned_bytes` doesn't work. A satisfying answer is more subtle than just that, though.

I go with `from_compressed_bytes` as is now, but hand off the conversion from assigned bytes to bits to a helper function. This helper function can also be moved to a utils module.

This makes the footprint of `from_compressed_bytes` lighter; also note: I am already witnessing the bytes of the `x`coordinate (and do not compute it in-circuit).

zk_stdlib/examples/cardano_signature.rs
remalue: #348

zk_stdlib/examples/cardano_signature.rs
```diff
@@ -0,0 +1,319 @@
...
+//! [libsodium](https://github.com/jedisct1/libsodium) uses the following verification criteria in
+//! `crypto_sign/ed25519/ref10/open.c`:
+//!   * cofactorless verification equation R = s * B - h * A,
+//!   * canonicity checks for bytes of s, A,
+//!   * subgroup-checks for R, A.
+//!
+//! This example uses the following verification criteria:
+//!   * cofactorless verification R = s * B - h * A,
+//!   * in-circuit canonicity checks for bytes of s, R, A,
+//!   * in-circuit subgroup-check for R, A.
```
miguel-ambrona, line 42: Since `A` is a public input, can we avoid the canonicity checks and subgroup-checks on `A`?

zkir/src/zkir.rs
miguel-ambrona, line 1: The `zkir` crate was removed. Please, rebase and remove this file.

zk_stdlib/examples/cardano_signature.rs
remalue: I didn't find a way without declaring as public input (which would change the relation) or changing the implementation of compression/decompression. Maybe this is another drawback of this implementation.

miguel-ambrona: What do you mean by "without declaring as public input"?

It is a public input of your relation, right? You don't have a choice.

miguel-ambrona (approved):

PR #311, opened 2026-04-06
Multi-circuit aggregator

miguel-ambrona:
Introduces a multi-circuit aggregator that folds proofs from different inner circuits (different VKs) into a single IVC proof. Inner circuits can be set up and proved on-the-fly, after the aggregator is initialized. The only requirement is a shared  architecture, circuit size, and SRS.

We may be able to relax the circuit size condition in the future.


aggregation/src/multi_circuit_aggregator/mod.rs
```diff
@@ -0,0 +1,79 @@
+//! Multi-circuit proof aggregation via IVC.
+//!
+//! This module provides an IVC-based proof aggregator that can aggregate proofs
+//! from different inner circuits (i.e. circuits with different verifying keys)
+//! into a single succinct proof.
+//!
+//! All inner circuits must share the same constraint system and evaluation
+//! domain. They differ only in their verifying keys (fixed commitments).
```
iquerejeta, line 8: Why must they share the evaluation domain?

aggregation/src/multi_circuit_aggregator/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+//!    recomputed from the provided list of claims (decider check).
+//!    c. The accumulated inner-proof verification passes the pairing check
+//!    (decider check).
+//!
+//! 2. Check that the aggregated claims are acceptable. Step 1 guarantees that
+//!    every claim has a valid inner proof, but says nothing about *what* was
+//!    proved. It is up to the verifier to decide whether the claims are
+//!    meaningful by checking that each VK belongs to a trusted circuit, whose
+//!    setup was run by the verifier and whose architecture meets the
+//!    domain-specific requirements.
```
iquerejeta, line 42: I don't understand this comment: 

```
and whose architecture meets the domain-specific requirements.
```

aggregation/src/multi_circuit_aggregator/utils.rs
```diff
@@ -0,0 +1,105 @@
...
+/// ([`assign_and_hash_vk`]).
+pub fn compute_vk_hash(vk: &MidnightVK) -> F {
+    let vk = vk.vk();
+    let to_raw = Hashable::<PoseidonState<F>>::to_input;
+
+    let vk_repr = vec![vk.transcript_repr()];
+    let fixed_coms: Vec<F> = vk.fixed_commitments().iter().flat_map(to_raw).collect();
+    let perm_coms: Vec<F> = vk.permutation().commitments().iter().flat_map(to_raw).collect();
+
+    <PoseidonChip<F> as HashCPU<F, F>>::hash(&[vk_repr, fixed_coms, perm_coms].concat())
```
iquerejeta, line 49: The fixed/perm commitments are already part of the `transcript_repr()`. Isn't `transcript_repr()` precisely what you want here?

aggregation/src/multi_circuit_aggregator/claims.rs
```diff
@@ -0,0 +1,69 @@
+//! Claims and supporting traits for multi-circuit aggregation.
+//!
+//! A [`Claim`] is a `(vk, statement)` pair representing the assertion that
+//! `statement` holds under the inner circuit identified by `vk`. The aggregator
+//! collects one claim per IVC step.
+//!
+//! Because different inner circuits have different statement types, statements
+//! are stored as `Box<dyn Statement>` for type-erased access. The
```
iquerejeta, line 8: Don't really understand what this means: 

```
are stored as `Box<dyn Statement>` for type-erased access
```

it may be because I'm not familiar with 'type-erased access' concept.

aggregation/src/multi_circuit_aggregator/claims.rs
```diff
@@ -0,0 +1,69 @@
...
+//! and [`TypedStatement`] bridges it to the [`Statement`] trait object.
+
+use std::fmt::Debug;
+
+use midnight_zk_stdlib::MidnightVK;
+
+use super::AggregableRelation;
+use crate::ivc::F;
+
+/// A claim: an inner-circuit verifying key paired with a corresponding
```
iquerejeta, line 19:
```suggestion
/// An inner-circuit verifying key paired with a corresponding
```

iquerejeta (commented):
Very nice interface - I don't fully understand the limitation of forcing the same K. Other than that, very minor comments.

aggregation/src/multi_circuit_aggregator/utils.rs
miguel-ambrona: Unfortunately `transcript_repr` is not enough. The reason is that computing the `transcript_repr` in-circuit is really expensive as it also involves the constrain system and other things.

Note that this must be computed in-circuit, because it is the way we are constraining the (otherwise free) witnessed vk bases.

aggregation/src/multi_circuit_aggregator/mod.rs
miguel-ambrona: I'll rephrase it. It is not very deep, I just want to clarify that there is a constraint on the architecture of circuits that can be aggregated.

aggregation/src/multi_circuit_aggregator/claims.rs
miguel-ambrona: I'll rephrase it.

aggregation/src/multi_circuit_aggregator/mod.rs
miguel-ambrona: Because our verifier has a hard-coded "omega". This can be generalized with some cost.

iquerejeta (approved):

PR #312, opened 2026-04-07
Pass on LogUp protocol

miguel-ambrona:
Audit pass on the LogUp lookup argument: reviewed documentation, renamed variables for clarity, and revisited some documentation comments.

- Increase blinding factors to account for logup helper polynomial evaluations (Fixes midnightntwrk/midnight-security#1)
- Add blinds to the multiplicities polynomial on non-usable rows (previously filled with zeros, leaking the lookup pattern)


iquerejeta (approved):
Great finding - these ZK bugs are always hard to catch 🥲

Issue #243, opened 2026-03-03
Simplify `CommitmentReference`


miguel-ambrona:
Cases `Chopped` and `Linear` could be shared as explained [here](https://github.com/midnightntwrk/midnight-zk/pull/190#discussion_r2845655129).

miguel-ambrona:
Closed by https://github.com/midnightntwrk/midnight-zk/pull/314

PR #314, opened 2026-04-08
Remove `Chopped` case from `CommitmentReference`

miguel-ambrona:
Fixes https://github.com/midnightntwrk/midnight-zk/issues/243


remalue (approved):
Nice, LG!

I like that you removed the chopped case (and not the linear one), I found the checks in the other case weird.

Btw: `Chopped` and `from_parts` wasn't used anywhere (anymore)?

miguel-ambrona:
> Nice, LG!
> 
> I like that you removed the chopped case (and not the linear one), I found the checks in the other case weird.
> 
> Btw: `Chopped` and `from_parts` wasn't used anywhere (anymore)?

Yes, they were not used anymore, that's why I decided to simply drop `Chopped` instead of adapting one of the two cases to handle both.

PR #317, opened 2026-04-08
Add p256 to stdlib

iquerejeta:
Adds the P256 chip to std lib.


zk_stdlib/src/lib.rs
```diff
@@ -114,6 +115,9 @@ type NG = NativeGadget<F, P2RDecompositionChip<F>, NativeChip<F>>;
 type Secp256k1BaseChip = FieldChip<F, k256_mod::Fp, MEP, NG>;
 type Secp256k1ScalarChip = FieldChip<F, k256_mod::Fq, MEP, NG>;
 type Secp256k1Chip = ForeignWeierstrassEccChip<F, K256, MEP, Secp256k1ScalarChip, NG>;
+type P256BaseChip = FieldChip<F, p256_mod::Fp, MEP, NG>;
+type P256ScalarChip = FieldChip<F, p256_mod::Fq, MEP, NG>;
+type P256Chip = ForeignWeierstrassEccChip<F, P256, MEP, P256ScalarChip, NG>;
```
John-Gong-Math, line 120: Should we use `Secp256r1Chip` instead of `P256Chip`, as we did for `K256/Secp256k1`?

zk_stdlib/src/lib.rs
```diff
@@ -323,6 +337,13 @@ impl ZkStdLib {
             .map(|(curve_config, scalar_chip)| {
                 ForeignWeierstrassEccChip::new(curve_config, &native_gadget, scalar_chip)
             });
+        let p256_scalar_chip = (config.p256_scalar_config.as_ref())
+            .map(|scalar_config| FieldChip::new(scalar_config, &native_gadget));
+        let p256_curve_chip = (config.p256_config.as_ref()).zip(p256_scalar_chip.as_ref()).map(
+            |(curve_config, scalar_chip)| {
+                ForeignWeierstrassEccChip::new(curve_config, &native_gadget, scalar_chip)
+            },
+        );
```
John-Gong-Math, line 346: The same, shall we make the naming consistent with `secp256k1`?

John-Gong-Math (commented):
I am wondering which one is better in `stdlib`, `P256` or `Secp256r1`?

zk_stdlib/src/lib.rs
iquerejeta: I feel that P256 is the more widely used name, given that it is the one use in the standard. Maybe we could name the chip as Secp256r1, and the field to enable it P256. thoughts?

miguel-ambrona: I agree with Iñigo.

miguel-ambrona (dismissed):
LGTM!

It came to mind that the scalar chip (and the base chip) are available through the curve chip. For example: https://github.com/midnightntwrk/midnight-zk/blob/iquerejeta/p256_stdlib/zk_stdlib/examples/bitcoin_signature.rs#L69-L71

So it may be cleaner to only expose the curve chip in zk_stdlib and document it properly explaining it contains the other chips and how to get them. That would be cleaner. Not just for P256, but for the other existing curve chips.

John-Gong-Math (dismissed):

miguel-ambrona (approved):

PR #318, opened 2026-04-09
Avoid redundant selector conversion in fixed VK assignment

jtcoolen:
This change updates `assign_fixed_vk` to only convert selectors to fixed columns when selectors are present. For already-finalised constraint systems, it now preserves the existing representation instead of reprocessing it.

This keeps VK assignment behaviour consistent across assignment paths and avoids discrepancies caused by unnecessary selector normalisation.


miguel-ambrona:
Thanks, Julien. Can you explain a bit why you need this?

jtcoolen:
After rebasing a prototype recursive prover onto master (that hardcodes the vk of the child proofs as constant with `assign_fixed_vk`), noticed a host/circuit divergence in fixed-base indexing during recursive verification. In practice, the proof accumulator built in-circuit used a different fixed-key set (for example missing or shifted `<circuit_name>_vk_fixed_com_35..42` relative to host expectations), which made the returned accumulator differ and triggered circuit satisfiability and `dual_msm` check failures.

miguel-ambrona:
Can you provide a minimal example showcasing the issue you describe?

miguel-ambrona:
@jtcoolen Do you still need this?

jtcoolen:
@miguel-ambrona Yes we need it, this is needed to keep the constant VK assignment function in line with the variable VK assignment https://github.com/midnightntwrk/midnight-zk/commit/b76794bc0ef5683ef6fa6f6b65f7439be06cec2b#diff-d92f39472a5be9a308ada386c74177d9a1111b0d9b8edc27f18426b0b2ca2c42R191

miguel-ambrona:
I still do not understand why this can be necessary. Can you clarify it?

jtcoolen:
This is needed because `directly_convert_selectors_to_fixed` is not a pure/idempotent normalisation pass. Even when `cs.num_selectors() == 0`, calling it again on an already-finalised CS can still rewrite selector metadata, in particular the gate selector indices that were already remapped to fixed-column indices during the original finalisation. In recursive verification those indices are later used to derive fixed VK commitment names / fixed-base MSM terms.

As a result, `assign_fixed_vk` could produce an assigned VK whose fixed-key names differ from the host-side VK representation, leading to accumulator divergence and failures in satisfiability / `dual_msm`. The fix makes `assign_fixed_vk` behave like the variable/public VK assignment path: only perform selector-to-fixed conversion when selectors are actually present; otherwise preserve the existing finalised CS representation.

miguel-ambrona:
I will take care of this in a different way.

Thank you, @jtcoolen . I'll close this PR when it is dealt with.

PR #321, opened 2026-04-10
fix: replace assert_eq! with error return in public_inputs

Olexandr88:
A panic here is inconsistent with the rest of the function which uses ? and returns Result. Library code should propagate errors rather than panic on unexpected state.


miguel-ambrona:
Hi @Olexandr88 , thanks for your contribution.

Unfortunately, this is not the official ZKIR, it is just a prototype, which we are not going to maintain. I'm removing it:
https://github.com/midnightntwrk/midnight-zk/pull/323

Olexandr88:
Hi @miguel-ambrona Thanks for the clarification!

PR #322, opened 2026-04-10
Introduce `Blake2b256` transcript hash for on-chain verification

### LINKED ISSUE #299: Feat: have a plutus native hash function for the transcript
Feature Request: Allow for Blake2b digest of a smaller size to support Cardano Plutus verification. As I understand it, currently midnight requires [this](https://github.com/midnightntwrk/midnight-zk/blob/e2e2892004ee53b9c40688c28b48907be1095af8/proofs/src/transcript/implementors.rs#L145)

```rust
        assert!(hash_output.len() >= (midnight_curves::Fq::NUM_BITS as usize / 8) + 12);
```
So the hash output length is minimally 44

## User Need:
A user might want to verify a circuit from this lib directly on Cardano (which supports BLS12381). Unfortunately, the only available hash functions are

- sha2_256
- sha3_256
- blake2b_224
- blake2b_256
- keccak_256
- ripemd_160

## Expected Benefit:
Verify midnight halo2 circuits in plutus.

## Details:
I think that supporting Blake2b_224 would be best, as it is the cheapest to run on-chain, but also as the size of the digest directly fits in the field (255 bits). That said, I am unsure if this is bad for biasing.

I also understand that this change might only be warranted if other technical requirements of on-chain verification are clear. For example, if it turns out that the current verification can never run on-chain, adding this is useless.

miguel-ambrona:
Fixes https://github.com/midnightntwrk/midnight-zk/pull/299.


perturbing (approved):
LGTM, thank you!

iquerejeta (approved):

PR #331, opened 2026-04-14
Passport verification (2/5) – library asjustments

### LINKED ISSUE #343: Small library adjustments
Expose missing features for the passport PoC. Includes non-exposed variable-length hashes, and handle the Stack overflow issue of Vector Gadget.

irakoton:
Small PR to add missing features required for the passport verification.
1. Adding big endian conversions for biguint
2. Exposing varlen hash gadgets (not available before for, e.g., Poseidon)
3. Solve VectorGadget StackOverflow issue by boxing the buffer

### DIFF
--- a/circuits/src/biguint/biguint_gadget.rs
+++ b/circuits/src/biguint/biguint_gadget.rs
@@ -562,7 +562,7 @@ where
     }
 
     /// Returns a vector of assigned bits representing the given assigned big
-    /// unsigned integer little-endian.
+    /// unsigned integer in little-endian order.
     pub fn to_le_bits(
         &self,
         layouter: &mut impl Layouter<F>,
@@ -589,8 +589,22 @@ where
         Ok(bits)
     }
 
+    /// Returns a vector of assigned bits representing the given assigned big
+    /// unsigned integer in big-endian order. Delegates to
+    /// [`to_le_bits`](Self::to_le_bits) and reverses the result.
+    pub fn to_be_bits(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        x: &AssignedBigUint<F>,
+    ) -> Result<Vec<AssignedBit<F>>, Error> {
+        let mut bits = self.to_le_bits(layouter, x)?;
+        bits.reverse();
+
+        Ok(bits)
+    }
+
     /// Returns a vector of assigned bytes representing the given assigned big
-    /// unsigned integer little-endian.
+    /// unsigned integer in little-endian order.
     #[allow(clippy::assertions_on_constants)]
     pub fn to_le_bytes(
         &self,
@@ -615,8 +629,22 @@ where
         Ok(bytes)
     }
 
+    /// Returns a vector of assigned bytes representing the given assigned big
+    /// unsigned integer in big-endian order. Delegates to
+    /// [`to_le_bytes`](Self::to_le_bytes) and reverses the result.
+    #[allow(clippy::assertions_on_constants)]
+    pub fn to_be_bytes(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        x: &AssignedBigUint<F>,
+    ) -> Result<Vec<AssignedByte<F>>, Error> {
+        let mut bytes = self.to_le_bytes(layouter, x)?;
+        bytes.reverse();
+        Ok(bytes)
+    }
+
     /// Returns the assigned big unsigned integer represented by the given
-    /// vector of assigned bits, by interpreting it in little-endian.
+    /// vector of assigned bits, by interpreting it in little-endian order.
     pub fn from_le_bits(
         &self,
         layouter: &mut impl Layouter<F>,
@@ -639,7 +667,21 @@ where
     }
 
     /// Returns the assigned big unsigned integer represented by the given
-    /// vector of assigned bytes, by interpreting it in little-endian.
+    /// vector of assigned bits, by interpreting it in big-endian order.
+    /// Delegates to [`from_le_bits`](Self::from_le_bits) with the input
+    /// reversed.
+    pub fn from_be_bits(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        bits: &[AssignedBit<F>],
+    ) -> Result<AssignedBigUint<F>, Error> {
+        let mut bits = bits.to_vec();
+        bits.reverse();
+        self.from_le_bits(layouter, &bits)
+    }
+
+    /// Returns the assigned big unsigned integer represented by the given
+    /// vector of assigned bytes, by interpreting it in little-endian order.
     #[allow(clippy::assertions_on_constants)]
     pub fn from_le_bytes(
         &self,
@@ -663,6 +705,21 @@ where
             limbs,
             limb_size_bounds,
         })
+    }
+
+    /// Returns the assigned big unsigned integer represented by the given
+    /// vector of assigned bytes, by interpreting it in big-endian order.
+    /// Delegates to [`from_le_bytes`](Self::from_le_bytes) with the input
+    /// reversed.
+    #[allow(clippy::assertions_on_constants)]
+    pub fn from_be_bytes(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        bytes: &[AssignedByte<F>],
+    ) -> Result<AssignedBigUint<F>, Error> {
+        let mut bytes = bytes.to_vec();
+        bytes.reverse();
+        self.from_le_bytes(layouter, &bytes)
     }
 
     /// Returns `1` iff `x < y`.
--- a/circuits/src/biguint/biguint_gadget.rs
+++ b/circuits/src/biguint/biguint_gadget.rs
@@ -590,8 +590,7 @@ where
     }
 
     /// Returns a vector of assigned bits representing the given assigned big
-    /// unsigned integer in big-endian order. Delegates to
-    /// [`to_le_bits`](Self::to_le_bits) and reverses the result.
+    /// unsigned integer in big-endian order.
     pub fn to_be_bits(
         &self,
         layouter: &mut impl Layouter<F>,
@@ -599,7 +598,6 @@ where
     ) -> Result<Vec<AssignedBit<F>>, Error> {
         let mut bits = self.to_le_bits(layouter, x)?;
         bits.reverse();
-
         Ok(bits)
     }
 
@@ -630,8 +628,7 @@ where
     }
 
     /// Returns a vector of assigned bytes representing the given assigned big
-    /// unsigned integer in big-endian order. Delegates to
-    /// [`to_le_bytes`](Self::to_le_bytes) and reverses the result.
+    /// unsigned integer in big-endian order.
     #[allow(clippy::assertions_on_constants)]
     pub fn to_be_bytes(
         &self,
@@ -668,8 +665,6 @@ where
 
     /// Returns the assigned big unsigned integer represented by the given
     /// vector of assigned bits, by interpreting it in big-endian order.
-    /// Delegates to [`from_le_bits`](Self::from_le_bits) with the input
-    /// reversed.
     pub fn from_be_bits(
         &self,
         layouter: &mut impl Layouter<F>,
@@ -709,8 +704,6 @@ where
 
     /// Returns the assigned big unsigned integer represented by the given
     /// vector of assigned bytes, by interpreting it in big-endian order.
-    /// Delegates to [`from_le_bytes`](Self::from_le_bytes) with the input
-    /// reversed.
     #[allow(clippy::assertions_on_constants)]
     pub fn from_be_bytes(
         &self,
--- a/circuits/src/hash/poseidon/poseidon_varlen.rs
+++ b/circuits/src/hash/poseidon/poseidon_varlen.rs
@@ -94,37 +94,12 @@ impl<F: PoseidonField> VarLenPoseidonGadget<F> {
         Ok(result)
     }
 
-    /// Format the last chunk of data so it is zeroed after the effective
-    /// payload. Given chunk = [x1, x2, ..., xn], with n = RATE, returns
-    /// [x1, ..., x_{offset-1}, 0, ..., 0]. If offset = 0, the chunk is
-    /// returned intact.
-    fn constrain_last_chunk(
-        &self,
-        layouter: &mut impl Layouter<F>,
-        chunk: &[AssignedNative<F>],
-        offset: &AssignedNative<F>,
-    ) -> Result<Vec<AssignedNative<F>>, Error> {
-        assert_eq!(chunk.len(), RATE);
-        let ng = &self.native_gadget;
-
-        let mut chunk = chunk.to_vec();
-        let zero = ng.assign_fixed(layouter, F::ZERO)?;
-        let mut after_data: AssignedBit<F> = ng.assign_fixed(layouter, false)?;
-        for (i, elem) in chunk.iter_mut().enumerate().skip(1) {
-            let b = ng.is_equal_to_fixed(layouter, offset, F::from(i as u64))?;
-            after_data = ng.xor(layouter, &[b, after_data])?;
-            *elem = ng.select(layouter, &after_data, &zero, elem)?;
-        }
-
-        Ok(chunk)
-    }
-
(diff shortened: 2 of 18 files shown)


circuits/src/biguint/biguint_gadget.rs
```diff
@@ -589,8 +589,22 @@ where
         Ok(bits)
     }
 
+    /// Returns a vector of assigned bits representing the given assigned big
+    /// unsigned integer in big-endian order. Delegates to
+    /// [`to_le_bits`](Self::to_le_bits) and reverses the result.
```
miguel-ambrona, line 594: I would not mention the fact that it delegates to `to_le_bits` in the doc.
Maybe under `//` but not `///`.

circuits/src/biguint/biguint_gadget.rs
```diff
@@ -615,8 +629,22 @@ where
         Ok(bytes)
     }
 
+    /// Returns a vector of assigned bytes representing the given assigned big
+    /// unsigned integer in big-endian order. Delegates to
+    /// [`to_le_bytes`](Self::to_le_bytes) and reverses the result.
```
miguel-ambrona, line 634: Again, mind the details on the implementation through `to_le_bytes`.

circuits/src/biguint/biguint_gadget.rs
```diff
@@ -589,8 +589,22 @@ where
...
+    /// unsigned integer in big-endian order. Delegates to
+    /// [`to_le_bits`](Self::to_le_bits) and reverses the result.
+    pub fn to_be_bits(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        x: &AssignedBigUint<F>,
+    ) -> Result<Vec<AssignedBit<F>>, Error> {
+        let mut bits = self.to_le_bits(layouter, x)?;
+        bits.reverse();
+
```
miguel-ambrona, line 602: Mind this empty line which is not present in the `to_be_bytes` one.

circuits/src/hash/sha256/sha256_chip.rs
```diff
@@ -135,6 +135,16 @@ pub struct Sha256Chip<F: CircuitField> {
...
 }
 
+impl<F: CircuitField> Sha256Chip<F> {
+    /// Returns a [`VarLenSha256Gadget`] wrapping this chip for variable-length
+    /// hashing.
+    pub fn varlen_gadget(&self) -> super::VarLenSha256Gadget<F> {
+        super::VarLenSha256Gadget {
+            sha256chip: self.clone(),
+        }
+    }
```
miguel-ambrona, line 145: This function signature hints that it is "cheap" to call this function, however, every call clones self and builds a fresh `VarLenSha256Gadget`. I guess it is ok, but I wonder how we can prevent users from calling this all the time.

zk_stdlib/src/lib.rs
```diff
@@ -828,6 +834,22 @@ impl ZkStdLib {
             .hash(layouter, input)
     }
 
+    /// Variable-length Poseidon hash. Takes an [`AssignedVector`] of native
+    /// field elements with chunk alignment 2 (Poseidon rate) and maximum
+    /// length `M`. `M` must be a multiple of 2.
```
miguel-ambrona, line 839: Can we assert the condition of `M % 2 == 0`?

zk_stdlib/src/lib.rs
```diff
@@ -828,6 +834,22 @@ impl ZkStdLib {
...
+        &self,
+        layouter: &mut impl Layouter<F>,
+        input: &AssignedVector<F, AssignedNative<F>, M, 2>,
+    ) -> Result<AssignedNative<F>, Error> {
+        let poseidon_chip = self
+            .poseidon_gadget
+            .as_ref()
+            .unwrap_or_else(|| panic!("ZkStdLibArch must enable poseidon"));
+        let gadget = VarLenPoseidonGadget::new(poseidon_chip, &self.native_gadget);
+        gadget.poseidon_varlen(layouter, input)
```
miguel-ambrona, line 850: I would avoid creating a new gadget every time `ZkStdLib::poseidon_varlen` is called.

We can build an optional poseidon_varlen_gadget inside our struct and call it exactly as we call the poseidon_gadget.

zk_stdlib/src/lib.rs
```diff
@@ -872,6 +894,21 @@ impl ZkStdLib {
...
+    pub fn sha2_256_varlen<const M: usize>(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        input: &AssignedVector<F, AssignedByte<F>, M, 64>,
+    ) -> Result<[AssignedByte<F>; 32], Error> {
+        *self.used_sha2_256.borrow_mut() = true;
+        self.sha2_256_chip
+            .as_ref()
+            .expect("ZkStdLibArch must enable sha256")
+            .varlen_gadget()
```
miguel-ambrona, line 908: Same comment as for poseidon.
ZkStdLib should hold an optional SHA256 gadget which you query here.

Indeed, given that this is how you use the `.varlen_gadget()` method that you implemented for the `Sha256Chip`, I would remove such method as it will not be necessary.

circuits/src/hash/poseidon/poseidon_varlen.rs
```diff
@@ -165,13 +140,7 @@ impl<F: PoseidonField> VarLenPoseidonGadget<F> {
...
             updating = ng.xor(layouter, &[b, updating])?;
 
-            register = if i == MAX_LEN / RATE {
-                // Constrain vector filler values in the last chunk.
-                let last_chunk = self.constrain_last_chunk(layouter, chunk, &last_chunk_len)?;
-                self.cond_update(layouter, &register, &last_chunk, &updating)?
-            } else {
-                self.cond_update(layouter, &register, chunk, &updating)?
-            }
+            register = self.cond_update(layouter, &register, chunk, &updating)?;
```
miguel-ambrona, line 143: I'm not sure this is enough. We should discuss it with @davidnevadoc .

circuits/src/parsing/base64_chip.rs
```diff
@@ -233,10 +233,10 @@ impl<F: CircuitField, const M: usize, const A: usize> Base64VarInstructions<F, M
         self.native_gadget.assert_zero(layouter, &check)?;
 
         // Compute decoded buffer.
-        let out_buffer = self.decode_base64(layouter, &b64_input.0.buffer, true)?;
+        let out_buffer = self.decode_base64(layouter, &*b64_input.0.buffer, true)?;
```
miguel-ambrona, line 236: `&*` is weird

circuits/src/instructions/vector.rs
```diff
@@ -75,16 +79,19 @@ where
...
     /// padding with a 1, and the ones that represent payload data with a 0.
+    /// Also returns the (start, end) limits of the data in the buffer, since
+    /// they are computed internally.
+    #[allow(clippy::type_complexity)]
     fn padding_flag(
         &self,
         layouter: &mut impl Layouter<F>,
         input: &AssignedVector<F, T, M, A>,
-    ) -> Result<[AssignedBit<F>; M], Error>;
+    ) -> Result<(Box<[AssignedBit<F>; M]>, VectorBounds<F>), Error>;
```
miguel-ambrona, line 89: Why the `Box`?

circuits/src/hash/poseidon/poseidon_varlen.rs
iquerejeta, line 143: I'm a bit in doubt of this. In the 'cond_update' function, it says "chunk is expected to have length `RATE`. However, this is not guaranteed here. Moreover, here you may be using elements of the last chunk that are not part of the effective payload. Why are you no longer constraining the last chunk?

circuits/src/vec/vector_gadget.rs
```diff
@@ -121,28 +132,31 @@ where
...
-    ) -> Result<[AssignedBit<F>; M], Error> {
+    ) -> Result<(Box<[AssignedBit<F>; M]>, VectorBounds<F>), Error> {
         let ng = &self.native_gadget;
-        let (start, end) = self.get_limits(layouter, input)?;
+        let limits = self.get_limits(layouter, input)?;
+        let (start, end) = &limits;
         let mut is_data: AssignedBit<F> = ng.assign_fixed(layouter, true)?;
 
-        let result = (0..M - A)
+        let result = (0..M - A + 1)
```
iquerejeta, line 141: here you could use the `=` notation. I think it'll be clearer.

circuits/src/parsing/data_types.rs
```diff
@@ -103,60 +109,173 @@ where
...
+    /// specified format, returns `DD + 100 * MM + 10_000 * YYYY` as an
+    /// assigned native value. This function does not check the validity of
+    /// the date (e.g. `"32011990"` in DDMMYYYY is accepted as 32 January
+    /// 1990).
+    ///
+    /// For two-digit year formats ([`DateFormat::YYMMDD`]), `century_base`
+    /// must be `Some(N)` where N is an assigned value in [0, 99]. The year
+    /// is then resolved as `1900 + YY + (if YY < N { 100 } else { 0 })`,
+    /// i.e. the 100-year window is [1900+N, 2000+N). The caller is
+    /// responsible for constraining N < 100.
```
iquerejeta, line 121: I don't understand the century_base input. In particular this check `YY < N`. Why not have `Some(N) = century_base` and have the year resolved as `N * 100 + YY`. Our library is going to be used for centuries 😄

iquerejeta (commented):
The poseidon issue needs clarity.

circuits/src/biguint/biguint_gadget.rs
iquerejeta: Agree - I was doubting where to put this comment or not, but I would remove it as well.

circuits/src/instructions/vector.rs
irakoton: This is the patch avoiding stack overflows when working with big assigned vectors. Basically, the Box puts the buffer on the heap.

circuits/src/parsing/base64_chip.rs
irakoton: I agree. I couldn't find another way to make it type-check though. It's an issue related to Box, basically `*` is interpreted as "unbox" in this context, but the box itself can't be interpreted as a `&`.

zk_stdlib/src/lib.rs
irakoton: It is asserted in the VectorGadget creation functions (typically, `assign_with_filler`, see the `assert` circuits/src/vec/vector_gadget.rs L104). Do you think it's still necessary to add another check here?

circuits/src/hash/sha256/sha256_chip.rs
irakoton: I guess the simplest way would be to use `self` instead of `&self`. This way it's not misleading anymore, and the user will have to use `clone` themselves. Is it an ok solution?

circuits/src/parsing/data_types.rs
irakoton: The problem is we're at an intersection of centuries right now. There are old fossils like us who were born in 19XX, and some younger ones born in 20XX. So you kind of need to have a window covering all possible dates of birth, which isn't a fixed century [100N, 100N+99] window. And since the century base is, as far as I remember, used as a public input, you can't afford to have several century bases around at the same time.

From what I found, the online specs just say "you infer the century from context, good luck", and often go [1950, 2049] as of today. But at the same time I know people born before 1950, so we would need a different window today, and then switch to the standard [1950,2049] later when they're not around anymore... Hence the library allowing arbitrary windows without changing the VK.

If everything is still around in 60-70 years, I'll update the code to just do `100N + YY`, since it will be the right window then 😄

circuits/src/hash/poseidon/poseidon_varlen.rs
irakoton: It is guaranteed, it's another addition I made in this PR, namely, now the entry point of AssignedVector (e.g., assign_with_filler) now constrain that M is a multiple of A. @davidnevadoc can confirm, but I think this is the intended behaviour.

So here, this last chunk you're talking about shall not exist normally, the real last chunk is at index `MAX_LEN/RATE - 1`. I'm leaving this conversation open so that David may confirm.

irakoton: @iquerejeta had the same remark. I'm closing this conversation and we can continue on Inigo's side instead.

circuits/src/hash/sha256/sha256_chip.rs
miguel-ambrona: The solution is to remove this function entirely as I suggested in a comment below.

zk_stdlib/src/lib.rs
miguel-ambrona: I would add an assert. It is free anyway, since it only involves constants.

zk_stdlib/examples/identity/jwt/property_check.rs
```diff
@@ -1,18 +1,15 @@
...
 //!
-//! Unlike the automaton-based version, this circuit does **not** verify the
-//! structural correctness of the credential (e.g. valid JSON, UTF-8 encoding).
-//! It assumes the credential is well-formed and that field names are unique,
-//! which is guaranteed by the issuer whose signature has been verified.
-//! It also assumes there is no insignificant whitespace around `:` separators.
+//! This circuit does **not** verify the structural correctness of the
+//! credential (e.g. valid JSON, UTF-8 encoding). It assumes the credential
+//! is well-formed and that field names are unique, which is guaranteed by
+//! the issuer whose signature has been verified. It also assumes there is
```
remalue, line 11: Does this sentence mean, there should be no whitespace at all around `:`?

circuits/src/hash/poseidon/poseidon_varlen.rs
davidnevadoc: This doesn't look right, because I don't see how the filler at the end is being constrained.
This was previously done in `constrain_last_chunk` but I don't see the constrain now. 
I'll look more into it to confirm or deny this.

zk_stdlib/examples/identity/jwt/property_check.rs
irakoton: It's a bit more than that. It's a bit of a situation like the following. If I ask you to parse the following regex
```text
{\s*"name"\s*:\s*"[a-Z]+"\s*}`
```
(bracket, `"name"`, colon, the field content, bracket, and all those are separated by arbitrary number of spaces)
If you want to extract the field name, you may either parse the whole thing explicitly, i.e., check it has the correct regex structure, or just assume it does. In the latter case, you just take the field value as your witness `w` and check that `"name":"w"` is a substring of the input. Much lighter to verify, but requires to assume the credential has the right form, and that there is no space around the `:`.

Assumption of well-formedness are fine since the credentials are emitted by an authority whose signature has been verified in-circuit. And the no-space assumption is more debatable, but is how credentials are often issued in practice.

circuits/src/hash/poseidon/poseidon_varlen.rs
irakoton: That would be helpful if you could double check. My understanding was that the filler at the end was constrained by the "cond_" part of `cond_update`, and that what I deleted is whatever happens after `(M / A) * A`. Please double check, but I had the impression that the chunk you have in mind (the `payload length % A` trailing fillers after the payload) is the chunk `MAX_LEN / RATE - 1`, not `MAX_LEN / RATE`.

davidnevadoc: Confirmed. The last chunk may have unconstrained filler and therefore, the last chunk needs to be handled a bit differently. 
We can either restore constrain_last_chunk, or follow the approach of varlen SHA. Which is simply to go through all the last chunk overwriting all filler values with zeroes. Then, the last chunk can be  treated as all the rest.

I think we may be talking about different issues. I got a bit lost regarding the chunk number.
The bottom line is that we have M / A chunks. The loop must do the last one differently to handle trailing fillers right. The filler at the beginning comes in whole chunks and is dealt with by the `cond_update` mechanism.

iquerejeta (approved):
LGTM

PR #332, opened 2026-04-14
Passport verification (3/5) – varlen scanner chip

### LINKED ISSUE #344: Variable-length tools
Implement variable-length automaton parsing and substring checks.

irakoton:
Implements variable length variants for the scanner chip (automaton parsing and substring checks).
1. Variable length substring checks are not overly technical: since the existing substring checks already support 256 padding, it suffices to use 256 as a padding in the AssignedVector buffer, and the existing substring checks work quasi identically as now. The only difference is the index computation which has to be offset to take the padding into account.
2. For automata, we extend the transition tables by self loop with 256 on initial and final states. This handles automata naturally on 256 padded assigned vectors.


circuits/src/parsing/scanner/varlen.rs
```diff
@@ -0,0 +1,203 @@
...
+//!  - **Padding flags and limits are cached** and available at no extra cost
+//!    after construction.
+//!
+//! The chunk size `A` of the source [`AssignedVector`] determines filler
+//! placement and is preserved in the type.
+//!
+//! These properties make [`ScannerVec`] safe for use in both automaton parsing
+//! ([`ScannerChip::parse_varlen`](super::ScannerChip::parse_varlen)) and
+//! variable-length substring checks
+//! ([`ScannerChip::check_bytes_varlen`](super::ScannerChip::check_bytes_varlen)).
```
iquerejeta, line 35: These docs seem to be a better fit in the documentation of the structure, rather than the module.

circuits/src/parsing/scanner/varlen.rs
```diff
@@ -0,0 +1,203 @@
...
+/// original [`AssignedVector`] is preserved in the type for alignment
+/// purposes. The padding flags and limits are computed once at construction
+/// time and cached.
+#[derive(Debug, Clone)]
+pub struct ScannerVec<F: CircuitField, const M: usize, const A: usize> {
+    /// The effective length of the payload (constrained during construction).
+    length: AssignedNative<F>,
+    /// Buffer with filler positions constrained to 256. Boxed to keep
+    /// large buffers off the stack.
+    pub(crate) buffer: Box<[AssignedNative<F>; M]>,
```
iquerejeta, line 67: Why use `AssignedNative` and not `AssignedByte`? Like that you use the typesystem to avoid forgetting bound checks.

circuits/src/parsing/scanner/varlen.rs
```diff
@@ -0,0 +1,203 @@
...
+/// A variable-length vector with filler elements constrained to
+/// [`ALPHABET_MAX_SIZE`](super::ALPHABET_MAX_SIZE).
+///
+/// Stores a promoted buffer where filler positions are constrained to 256,
+/// along with the effective payload length. The chunk size `A` of the
+/// original [`AssignedVector`] is preserved in the type for alignment
+/// purposes. The padding flags and limits are computed once at construction
+/// time and cached.
+#[derive(Debug, Clone)]
+pub struct ScannerVec<F: CircuitField, const M: usize, const A: usize> {
```
iquerejeta, line 62: Can you use an `AssignedVector` (enforcing they inner type to be `AssignedByte`) instead of having `length` and `buffer` fields?

circuits/src/parsing/scanner/varlen.rs
```diff
@@ -0,0 +1,203 @@
...
+    pub fn padding_flags(&self) -> &[AssignedBit<F>; M] {
+        &self.padding_flags
+    }
+
+    /// Returns the effective length of the payload.
+    ///
+    /// # Cost
+    ///
+    /// Free: the value was computed and constrained during construction and
+    /// cached.
```
iquerejeta, line 99: It is weird to have this cost here - any reason why we are adding it to the docs/

circuits/src/parsing/scanner/varlen.rs
```diff
@@ -0,0 +1,203 @@
...
+        let (padding_flags, limits) = self.vector_gadget.padding_flag(layouter, &vec)?;
+
+        // Constrain filler positions to ALPHABET_MAX_SIZE.
+        let filler =
+            self.native_gadget.assign_fixed(layouter, F::from(ALPHABET_MAX_SIZE as u64))?;
+        let buffer: Box<[AssignedNative<F>; M]> = Box::new(
+            (vec.buffer.iter().zip(padding_flags.iter()))
+                .map(|(elem, is_padding)| {
+                    let native_elem = AssignedNative::from(elem);
+                    self.native_gadget.select(layouter, is_padding, &filler, &native_elem)
```
iquerejeta, line 189: I would keep the type `AssignedByte` when constructing the ScannerVec

circuits/src/parsing/scanner/automaton_chip.rs
```diff
@@ -117,11 +118,9 @@ where
     fn next_transition(
         &self,
         state: &AssignedNative<F>,
-        letter: &AssignedByte<F>,
+        letter: &AssignedNative<F>,
     ) -> Result<(Value<F>, Value<F>), Error> {
-        let letter_native: AssignedNative<F> = letter.into();
-        let target_opt =
-            state.value().zip(letter_native.value()).map(|(s, l)| self.get_transition(s, l));
+        let target_opt = state.value().zip(letter.value()).map(|(s, l)| self.get_transition(s, l));
```
iquerejeta, line 123: The documentation states that inputs are guaranteed to be bytes, and that final states use `ALPHABET_MAX_SIZE`, which is now used also for filler values. 

Does the final state need to be redefined to not colude with the filler values? Also, update the documentation accordingly:

```
//! Final states are encoded as dummy transitions labelled with the invalid
//! byte 256 (= `ALPHABET_MAX_SIZE`), pointing to state 0 with output 0.
//! Since input bytes are range-checked to `[0, 255]`, these transitions can
//! only be triggered explicitly by [`ScannerChip::assert_final_state`].
```

iquerejeta: I'm wondering if there are design patters where we can avoid loosing the information on the input. Maybe with an `enum`, it being a value, or padding.

circuits/src/parsing/scanner/substring_varlen.rs
```diff
@@ -0,0 +1,385 @@
...
+            });
+
+        Ok(())
+    }
+
+    /// Asserts that two byte slices of variable length are element-wise equal.
+    /// Proceeds by testing equality of the lengths, and checks that `v2` is a
+    /// substring of `v1` at index 0. If either of the two vectors is also used
+    /// as the `sequence` argument of another substring check, putting it as the
+    /// `v1` argument is the most cost efficient.
```
iquerejeta, line 118: Why use `check_bytes_varlen` instead of just asserting equality of each element (including the padding)?

circuits/src/parsing/scanner/substring_varlen.rs
```diff
@@ -0,0 +1,385 @@
...
+        v2: &ScannerVec<F, M2, A2>,
+    ) -> Result<(), Error> {
+        self.native_gadget.assert_equal(layouter, v1.len(), v2.len())?;
+        let zero: AssignedNative<F> = self.native_gadget.assign_fixed(layouter, F::ZERO)?;
+        self.check_bytes_varlen(layouter, v1, &zero, v2)
+    }
+
+    /// Asserts that two byte slices of variable and fixed length, respectively,
+    /// are element-wise equal. Proceeds by testing equality of the lengths,
+    /// and checks that `v2` is a substring of `v1` at index 0.
```
iquerejeta, line 137: Again, you are already giving away the size of the assigned vector, so you could do assert_equal over the effective payload.

iquerejeta (commented):
Good work - very nice organisation of the commit-history.

circuits/src/parsing/scanner/varlen.rs
irakoton: I could theoretically have an assigned vector to avoid having the two fields, but I could not enforce the AssignedByte inner type. Same issue as here, I actually need the fillers of the buffer to be constrained as non-bytes, so it's a bit unavoidable.

circuits/src/parsing/scanner/varlen.rs
irakoton: It's the same issue as a previous comment, I actively need for correctness that the filler positions are not bytes. My approach in the code is to have the entry sites guarantee the payload is AssignedBytes, that the filler is 256, and ScannerVec don't implement any length modifying function, to maintain this invariant.

Still, the range check is not completely lost, in the sense that you can retrieve an AssignedVector containing AssignedByte with `scanner_vec_to_byte_vector`, in case you really need. Typically happens in the passport when parsing something and hashing it later, IIRC.

circuits/src/parsing/scanner/substring_varlen.rs
irakoton: The two vectors don't have the same parameters M and A

circuits/src/parsing/scanner/substring_varlen.rs
irakoton: Indeed here

circuits/src/parsing/scanner/substring_varlen.rs
irakoton: can still short-circuit when all parameters are identical.

circuits/src/parsing/scanner/automaton_chip.rs
irakoton: It's fine, because the dummy transitions are the only one pointing to the state `0`, and we guarantee by construction that no transition do so in the actual transition table. So the self loops only share the letter 256 with the dummies, but live in different worlds.

I thought about alternative designs at some points, like enums, but they were super invasive and heavy. Honestly, there's a bit of acrobatics sometimes, in particular when you want to retrieve a vector of AssignedByte, but what I have now is the more readable I managed to produce.

circuits/src/parsing/scanner/varlen.rs
irakoton: For the soundness of the chips, I actively need that the fillers are not bytes (to guarantee they don't interfere with the parsing basically)

iquerejeta (approved):

PR #333, opened 2026-04-14
Passport verification (4/5) – ASN.1 parser

### LINKED ISSUE #345: ASN.1 support
Implement a general support for specifying and parsing ASN.1 encodings in circuit, with support for extracting fixlen/varlen fields.

irakoton:
Adds a library for specifying arbitrary ASN.1 specs, and a parser that automatically generates the related constraints. This is the central PR of the wave.

Specifications are basically a sequence of fixlen, varlen, and TLV (Tag-Length-Value) blocks. If they are marked for extraction, they are validated to be part of the initial input through substring checks, and T-L blocks are validated by automata. The length constraints of TLV blocks are enforced by arithmetic constraints.


irakoton:
Put on hold for now. In the threat model of identity applications (an official issuer signing a credential + signature verification in-circuit), it seems reasonable to trust the issuer to emit correctly formed credentials. In particular, it is not necessary to parse exactly ASN.1 credentials (e.g., in biometric passports). It is sufficient to use substring checks to extract fields, assuming they can be identified by a field name/prefix that is unique (at least with overwhelming probability). This is in practice much cheaper.

Future applications may however operate under different threat models, where the well-formedness of credentials needs to be enforced in-circuit for security (see, e.g., [this paper](https://eprint.iacr.org/2026/227.pdf)). If Midnight plans to support such applications and they happen to use ASN.1 formats, this PR would need to be re-opened.

Some related resources (parsing in ZK beyond automata techniques): [Coral](https://brave.com/blog/coral/). May come in handy in the future if this whole "parsing needs to be strictly enforced" pandora box needs to be opened in this repo.

PR #334, opened 2026-04-14
Passport verification (5/5) – Passport PoC

### LINKED ISSUE #346: Passport PoC
Use the library tool to verify the full trust chain of a biometric passport in ZK, and check that the owner is 18+.

irakoton:
Implements a passport verification pipeline. Defines an Asn1Spec and parses it automatically, to then use the parsed data in a standard identity signature-verification pipeline.


irakoton:
Closed for the same reason as #333 (no use case to support ASN.1 parsing for now). The passport PoC implemented in this PR should be re-implemented using more efficient substring checks instead of the ASN.1 tools.

PR #336, opened 2026-04-14
Optimizations on foreign Edwards chip

### LINKED ISSUE #265: Optimize foreign arithmetic on twisted Edwards curves
# Current State
The in-circuit foreign arithmetic on twisted Edwards curves was added by #241 and #242. The implementation of `double` falls back to `add`. Furthermore, the MSM implemented through `msm_by_bounded_scalars` and the scalar multiplication in `mul_by_constant` use a very basic form of double-and-add.

# Suggested Optimizations
Add the following optimizations to the foreign Edwards chip:
* Optimize the doubling operation.
* Optimize the MSM (e.g., by the windowed method, filtering scalars that are `0` or `1`, batching equal points).
* Optimize the custom gate identity `coord`.

### LINKED ISSUE #315: Dynamic lookups for Curve25519 MSM
Linked to #305

miguel-ambrona:
The goldenfiles diff speaks for itself.

image

Closes https://github.com/midnightntwrk/midnight-zk/issues/265
Closes https://github.com/midnightntwrk/midnight-zk/issues/315


circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -742,26 +901,23 @@ where
...
-                self.windowed_scalar_mul::<1>(layouter, &bits, b)?
-            } else if *num_bits <= 32 {
-                self.windowed_scalar_mul::<2>(layouter, &bits, b)?
-            } else {
-                self.windowed_scalar_mul::<3>(layouter, &bits, b)?
-            };
-            res = self.add(layouter, &res, &term)?;
-        }
+        // Decompose scalars to bits with tight bound, then chunk into windows.
+        const WS: usize = 4;
```
davidnevadoc, line 909: Is this the best size for full-sized scalars? Is it worth it to make it depend on the bound?

Following this line of reasoning, I think making different sized windows per point in the MSM is not worth the complexity. But maybe it is for single scalar muls. Is there a reasonable use of single multiplications with small bounded scalars where it may be worth it adjusting window size?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -779,22 +935,30 @@ where
...
+                }
             }
             // The doubling in the last iteration is not needed
             if i + 1 < scalar_bits.len() {
                 p = self.double(layouter, &p)?;
             }
         }
 
-        Ok(res)
+        Ok(res.unwrap_or(self.assign_fixed(layouter, C::CryptographicGroup::identity())?))
```
davidnevadoc, line 961: This gets eagerly evaluated. Is that intended or would it be better `unwrap_or_else`?
In practice, the cache make both equivalent if I'm not mistaken.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -779,22 +935,30 @@ where
...
         let scalar_bits = scalar.to_bits_le(None);
         let mut p = base.clone();
-        let mut res = self.assign_fixed(layouter, C::CryptographicGroup::identity())?;
+        let mut res = None;
 
         // Simple double-and-add
         for (i, b) in scalar_bits.iter().enumerate() {
             if *b {
-                res = self.add(layouter, &res, &p)?;
+                res = match res {
```
davidnevadoc, line 950: Nice saving!

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -861,51 +1025,114 @@ where
...
+            table.push(entry);
         }
         Ok(PrecomputedTable { table })
     }
 
-    /// Selects `table[w]` where `w` is the LE value of `bits`, using a binary
-    /// mux tree. Cost: 2^WS - 1 point selects.
-    fn multi_select<const WS: usize>(
+    /// Delegates to [`fill_dynamic_lookup_row`] with this chip's columns.
+    #[allow(clippy::type_complexity)]
```
davidnevadoc, line 1051: Thinking of an elegant way of removing the need for this 🤔

davidnevadoc (commented):
WIP

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -763,20 +763,23 @@ where
...
         let scalar_bits = scalar.to_bits_le(None);
         let mut p = base.clone();
-        let mut res = self.assign_fixed(layouter, C::CryptographicGroup::identity())?;
+        let mut res = None;
 
         // Simple double-and-add
         for (i, b) in scalar_bits.iter().enumerate() {
             if *b {
-                res = self.add(layouter, &res, &p)?;
+                res = match res {
```
remalue, line 771: Row count decreases because you save the addition `0 + p`?

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -70,7 +73,10 @@ where
 {
     base_field_config: FieldChipConfig,
     addition_config: AdditionConfig<C>,
-    _marker: PhantomData<C>,
+    // Dynamic lookup columns for windowed MSM table selection.
+    q_multi_select: Selector,
+    idx_col_multi_select: Column<Advice>,
+    tag_col_multi_select: Column<Fixed>,
```
davidnevadoc, line 79: Would be nice to have a mechanism to protect this column against misuse by other chips. Even if it is always initialized in `configure()` it can still be passed to another chip.

davidnevadoc (dismissed):
LGTM!

circuits/goldenfiles/cost-model.json
```diff
@@ -375,8 +375,8 @@
       "max_deg": "5",
       "permutations": "14",
       "point_sets": "5",
-      "rows": "159",
-      "table_rows": "385"
+      "rows": "138",
```
remalue, line 378: Hmm...I am unsure how to interpret this commit and what are the actual improvements. From the goldenfiles it seems you offload work to lookups. Is that correct? Is this commit worth the savings?

circuits/src/ecc/foreign/weierstrass_chip.rs
remalue, line 1: Maybe moving to `common.rs` and deleting from `weierstrass_chip.rs` would have been better in a separate  commit.

remalue:
First quick pass done. Now I need to have a look at the dynamic lookup technique.

circuits/src/ecc/foreign/common.rs
```diff
@@ -175,3 +180,117 @@ where
...
+        let sel = meta.query_selector(q_multi_select);
+        let not_sel = Expression::from(1) - sel.clone();
+
+        // This is a lookup of a column on itself!
+        //
+        // All identities are of the form: `(value, (1 - sel) * value)`.
+        // That means, every value in the column (it is actually 2 columns including the
+        // tag) is requested to be in the table! Most values, where the selector is
+        // disabled are trivially in the table (at least at their own position).
+        // Those values that we really wanna lookup have a `sel` value of 1, thus
```
remalue, line 214: "wanna lookup" -> "want to lookup"

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: `WS = 4` is optimal for most cases.

You are right that if the scalar bound is very small, smaller `WS` might be better.

I would not care about this for now. If at some point the efficiency of a single scalar mul becomes critical, we can revisit this. We may even want to implement fake GLV in that case.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: The cache will make it equivalent since the identity is for sure being assigned elsewhere.

I'll have a look at this and may switch to `unwrap_or_else`.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: Yes

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: How about a comment?

circuits/goldenfiles/cost-model.json
miguel-ambrona: IIUC, you are pointing to commit `b3b5c10c` titled _add sign flag to Edwards add gate_.

That saving does not come from lookups but from the fact that addition and doubling become cheaper, since one does not need to compute `1 + d * ...` and `1 - d * ...` prior to the call to the custom gate. It is enough to compute `d * ...` and the gate will take care of the rest.

circuits/src/ecc/foreign/weierstrass_chip.rs
miguel-ambrona: Why? If they are in the same commit you can compare in GitHub the code that was deleted to the one that was added.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -863,35 +875,92 @@ where
...
+
+    /// Loads a precomputed point table into the dynamic lookup.  Entry `i` is
+    /// paired with index `i` and the given `table_tag`.
+    fn load_multi_select_table(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        point_table: &[AssignedForeignEdwardsPoint<F, C, B>],
+        table_tag: F,
+    ) -> Result<(), Error> {
+        let indices: Vec<AssignedNative<F>> = (0..point_table.len())
```
remalue, line 911: Why do you assign the indices separately and not in the for loop?

circuits/src/ecc/foreign/common.rs
```diff
@@ -175,3 +180,117 @@ where
...
+            .chain(base_field_config.x_cols.iter())
+            .chain(base_field_config.z_cols.iter())
+            .map(|col| {
+                let val = meta.query_advice(*col, Rotation::cur());
+                (val.clone(), not_sel.clone() * val)
+            })
+            .collect::<Vec<_>>();
+
+        // Handle tag independently, since it is a fixed column
+        let tag = meta.query_fixed(tag_col_multi_select, Rotation::cur());
```
remalue, line 230: Wouldn't it suffice to enable the switch `(1-sel)` _only_ on the tag column?

circuits/src/ecc/foreign/common.rs
```diff
@@ -175,3 +180,117 @@ where
...
+        //
+        // All identities are of the form: `(value, (1 - sel) * value)`.
+        // That means, every value in the column (it is actually 2 columns including the
+        // tag) is requested to be in the table! Most values, where the selector is
+        // disabled are trivially in the table (at least at their own position).
+        // Those values that we really wanna lookup have a `sel` value of 1, thus
+        // `(1 - sel) * value` is zero and they themselves are not part of the lookup
+        // meaning they must be somewhere else (in the column positions that are part of
+        // the table). An exception is when value = 0 and tag = 0, but we actually use
+        // tag = 0 as a default case that we never want to lookup.
```
remalue, line 218: This comment (and the wider documentation of self-referential dynamic lookups) I found not very helpful. It basically repeats what is already written in the code.

How this type of lookup actually works (i.e., why it is sound) would help more here.

Can we take the opportunity and improve this doc?

circuits/src/ecc/foreign/common.rs
remalue: (not suggesting to do it, just trying to understand the minimal requirements for soundness)

circuits/goldenfiles/cost-model.json
remalue: Yes, exactly this commit. 

The question is: is addition really cheaper in the end? I see that the number of table queries increases, so I wasn't sure how to interpret this numbers.

circuits/src/ecc/foreign/weierstrass_chip.rs
remalue: It was feedback I got from the team myself.

circuits/goldenfiles/cost-model.json
miguel-ambrona: Ignore the `table_rows`. It is not representative of what's happening. It depends on the pow2range schedule. Also, note that it is global for the whole circuit, not local for the measured computation.
This commit did not introduce any lookup.

circuits/src/ecc/foreign/common.rs
miguel-ambrona: Can you elaborate? I do not understand your question.

circuits/src/ecc/foreign/common.rs
miguel-ambrona: How would you improve it? I find this comment very clear. It helps me a lot every time I revisit this part of the codebase.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -145,6 +170,8 @@ where
             native_gadget: native_gadget.clone(),
             base_field_chip,
             scalar_field_chip: scalar_field_chip.clone(),
+            tag_cnt: Rc::new(RefCell::new(1)),
```
remalue, line 173: This line is very important!

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -914,55 +1141,46 @@ where
...
+        let max_num_windows = *num_windows.iter().max().unwrap();
 
-        // Precompute tables for each base.
-        let tables: Vec<PrecomputedTable<F, C, B, WS>> = bases
-            .iter()
-            .map(|b| self.precompute::<WS>(layouter, b))
-            .collect::<Result<Vec<_>, _>>()?;
+        // Precompute tables for each base and load them into the dynamic lookup.
+        let tag_cnt = *self.tag_cnt.borrow();
+        self.tag_cnt.replace(tag_cnt + bases.len() as u64);
```
remalue, line 1150: Can this wrap to 0? Should we add a `debug_assert` for `!= 0` here?

Probably only a hypothetical scenario, and by no means practical.

remalue:
Very nice optimizations! 💪 Left some comments.

I found self-referential dynamic lookups clever (and a bit of evil magic).

Didn't see anything that bothered me. Will go over the lookups again in the afternoon, then I am good to approve.

circuits/src/ecc/foreign/common.rs
remalue: Some thoughts. But we can also discuss in a call, if you want.

* Comment on soundness (how is a break of soundness is prevented?).
* Clarify what the exceptional case of value = 0 (and tag = 0) means (it might break soundness, no?).
* Clarify what is actually looked up (how do you count 2 cols incl. the tag?).
* What does an identity of the form `(value, (1 - sel) * value)` mean?
* Why are only "most" of the values where sel = 0 in the table? Isn't sel = 0 the marker for table values and all of these values constitute the table (+0)?

I think the concept of _indexed_ lookups might help here (since the value(s) at a concrete table index is queried here).

circuits/src/ecc/foreign/common.rs
remalue: The identities `(v, (1-sel) * v)` are enforced on every column that takes part in the lookup, right?

Would it suffice to only enforce this on the tag column?

(And just have `(v, v)` on the other columns. The tag column alone should suffice to enforce the lookup query is into the table values)

miguel-ambrona: I still do not understand the question. What do you mean by "only enforce this on the tag column"?

Let me clarify the comment on how the self-referential lookup works and you may revisit this question again, please.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: This was considered long ago. We might have that kind of assertion in some places.

The actual check is not w.r.t. since the tag_cnt can only increase. The true check would be that the tag_cnt never wraps around the native field modulus. E.g. that F::NUM_BITS > 64.

remalue: No, I think you didn't understand.

`tag_cnt` is `u64`, right? According to doc, in release build it holds `u64::MAX + 1 == 0` . I am speaking of this wrap to 0, not around the native modulus.

circuits/src/ecc/foreign/common.rs
remalue: Self-referential lookups are of the form `T ⊆ (1-Sel)*T`, where `Sel` is the selector col, and `T` is one and the same column that stores table values and query values.

In the Edwards chip the lookup is of the form

`(Idx, X_cols, Y_cols, Tag) ⊆ ((1-Sel)*Idx, (1-Sel)*X_cols, (1-Sel)*Y_cols, (1-Sel)*Tag).`

Here `(A,B)` and `C * D` is a shorthand for `{(A(ω), B(ω)) : ω}` and `{C(ω) * D(ω): ω}`.

Does that make sense?

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: The ? makes it hard to use `unwrap_or_else` cleanly. I'll leave it as is.

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: We could introduce an alias for the output type, but I do not find it that complex anyway.

circuits/src/ecc/foreign/common.rs
miguel-ambrona: That makes sense, indeed, that's how the abuse of notation `T ⊆ (1-Sel)*T` should be interpreted.

What is your question?

miguel-ambrona: Are you suggesting to do this?

`(Idx, X_cols, Y_cols, Tag) ⊆ (Idx, X_cols, Y_cols, (1-Sel)*Tag)`?

circuits/src/ecc/foreign/edwards_chip.rs
miguel-ambrona: There is barely enough memory on Earth to build a circuit that overflows that `u64` and definitely our SRS could not deal with such circuit.

The real risk is what I mentioned, because the `tag_cnt` is ultimately converted to `F`, that it wraps-around the native field if the field is small (e.g. if we switched to a FRI-based PCS with small field).

circuits/src/ecc/foreign/common.rs
remalue: That's what I mean. But I am not suggesting to do it. I was trying to assess if this suffices for soundness.

circuits/src/ecc/foreign/edwards_chip.rs
remalue: > Probably only a hypothetical scenario, and by no means practical.

It was not about practicality. It was about making it explicit.

circuits/src/ecc/foreign/common.rs
miguel-ambrona: I think that could work, good idea. Let's explore this later today.

circuits/src/ecc/foreign/common.rs
```diff
@@ -205,17 +205,37 @@ pub(crate) fn configure_multi_select_lookup<F: CircuitField>(
...
-        // Those values that we really wanna lookup have a `sel` value of 1, thus
-        // `(1 - sel) * value` is zero and they themselves are not part of the lookup
-        // meaning they must be somewhere else (in the column positions that are part of
-        // the table). An exception is when value = 0 and tag = 0, but we actually use
-        // tag = 0 as a default case that we never want to lookup.
+        // Here, `value` is actually a tuple, but it is helpful to ignore this detail
+        // initially, for the sake of simplicity, we will come back to it later.
+        //
+        // The above should be interpreted as a set inclusion:
+        // {value(ω) | ω ∈ Ω}  ⊆  {(1 - sel(ω)) * value(ω) | ω ∈ Ω}.
```
remalue, line 215: This part is nice! I find it much clearer than Halo2's `lookup_any` syntax that was used before in the doc.

circuits/src/ecc/foreign/common.rs
```diff
@@ -205,17 +205,37 @@ pub(crate) fn configure_multi_select_lookup<F: CircuitField>(
...
+        //
+        // Observe that the table is then defined by all the entries where `sel = 0`,
+        // and there is no distinction between the intended lookup table and the values
+        // from other circuit regions where we simply want to skip this lookup check.
+        // This is not a problem (for soundness) because `value`, as we anticipated, is
+        // indeed a tuple `(value', tag)` where, in turn, `value'` may be a tuple
+        // itself. Importantly, `tag` is a fixed column which allows us to use it as a
+        // domain separator to solve the above ambiguity:
+        //
+        //  - The case `tag = 0` is reserved for disabled checks that are not supposed
```
remalue, line 233: Let's leave as is for now. And, if the discussion about removing the switch `(1-sel)` from other columns continues, revisit it. I have some thoughts on soundness here.

remalue (dismissed):

remalue (approved):

PR #350, opened 2026-04-17
Prover Optimization: affine SRS 

davidnevadoc:
Merges into  #349 

  ## Summary

  - Store the SRS points (`g`, `g_lagrange`) as `Vec<G1Affine>` instead of  `Vec<G1Projective>`.
 This eliminates the projective→affine batch conversion     that blst's Pippenger performs internally on every MSM call.
  - Add `G1Affine::multi_exp_affine` that calls blst's Pippenger directly on  affine points, and route `msm_specific` through it for BLS12-381.
  - Change `g_to_lagrange` to accept and return affine points.
  The FFT still  runs in projective coordinates internally, but the conversion is now encapsulated inside `g_to_lagrange`. This cleans up the codebase and its fine since we call this on SRS points that should always be in affine form.
  - Simplify SRS serialization: read/write `G1Affine` directly via `SerdeObject`, removing the `ProcessedSerdeObject` indirection for `G1` points. Format is unchanged (both old and new code serialize the compressed/uncompressed affine representation).
The serialization could still be cleaned up further, but that needs a deeper refactor in `proofs` and `curves`.
  - The `light_aggregator` stores its `lagrange_commitments` as projective (needed by the IPA). An affine→projective conversion is added at load time (not the hot path, so it does not hurt performance).

## Peformance improvement
| Optimization | k=15 (ms) | delta | k=17 (ms) | delta |
  | :--- | :-------: | :-----: | :-------: | :-----: |
  | Baseline  | 3214 | — | 13434 | — |
  | Affine SRS | 3149 | -2.0% | 13224 | -1.6% |

 **CPU**: Intel i5-11400H (6C/12T, 12 MB L3)
 Turbo disabled for stability

### DIFF
--- a/curves/src/bls12_381/g1.rs
+++ b/curves/src/bls12_381/g1.rs
@@ -422,8 +422,11 @@ impl G1Affine {
 impl G1Affine {
     /// Perform a multi-exponentiation on affine points.
     ///
-    /// `nbits` is the maximum number of significant bits across all scalars.
-    pub fn multi_exp_affine(points: &[Self], scalars: &[Fq], nbits: usize) -> G1Projective {
+    /// # Panics
+    ///
+    /// If |points| != |scalars|.
+    /// If |scalars| == 0 due to internal blst `mult` call.
+    pub fn multi_exp_affine(points: &[Self], scalars: &[Fq]) -> G1Projective {
         use blst::MultiPoint;
 
         let n = points.len();
@@ -439,7 +442,7 @@ impl G1Affine {
         }
 
         // Blst Pippenger MSM over affine points.
-        let res = affine_slice.mult(scalar_bytes.as_slice(), nbits);
+        let res = affine_slice.mult(scalar_bytes.as_slice(), 255);
         G1Projective(res)
     }
 }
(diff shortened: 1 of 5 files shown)


curves/src/bls12_381/g1.rs
```diff
@@ -419,6 +419,31 @@ impl G1Affine {
     }
 }
 
+impl G1Affine {
+    /// Perform a multi-exponentiation on affine points.
+    ///
+    /// `nbits` is the maximum number of significant bits across all scalars.
+    pub fn multi_exp_affine(points: &[Self], scalars: &[Fq], nbits: usize) -> G1Projective {
```
miguel-ambrona, line 426: Document that it expects `|points| = |scalars|` and panics if not?

miguel-ambrona: Document what happens if the scalars violate the given `nbits`?

proofs/src/utils/arithmetic.rs
```diff
@@ -31,22 +31,26 @@ where
...
         omega_inv = omega_inv.square();
     }
 
-    let mut g_lagrange = g_projective.to_vec();
-    best_fft(&mut g_lagrange, omega_inv, k);
-    parallelize(&mut g_lagrange, |g, _| {
+    // The FFT operates in projective coordinates (G1Affine is not closed
+    // under addition), so convert on entry and normalize back on exit.
+    let mut g_proj: Vec<C::CurveExt> = g_affine.iter().map(|p| (*p).into()).collect();
+    best_fft(&mut g_proj, omega_inv, k);
```
miguel-ambrona, line 44: Ok, this conversion is the "caveat" of having everything in affine, right? But from affine to projective it should be almost free. Correct?

miguel-ambrona (approved):
Very nice!

PR #351, opened 2026-04-17
Prover Optimization: Graph Evaluator

davidnevadoc:
Merges into #349 

  ## Summary

  Replace the interpreted `GraphEvaluator` (nested enum dispatch per step per   domain element) with `FlatGraphEvaluator`: a linear instruction sequence resolved at keygen time with pre-resolved buffer indices and bulk column loads.

Three commits:
  1. Flatten custom-gates evaluator into 7-opcode flat IR.
  2. Extend to lookup and trash evaluators; pre-compute cosets in parallel.  
  3. Add `evaluate_chunk<const BATCH>` — processes BATCH elements per graph      traversal, amortizing dispatch. Configurable via `FLAT_BATCH_SIZE` env var.
   
 ## TODO 
- [x]  Measure best batch size and hardcode it, instead of getting it from an environment variable.
- [x]  ~Merge #353 here~ Targets meta instead
 
 ## Performance improvement
| Commit | Description | k=15 bitcoin_sig (ms) | Δ vs base | k=17 cred_full (ms) | Δ vs base |
  |---|---|---:|---:|---:|---:|
  | `02e28d19` | baseline (main after CI update) | **3215.7** ± 8.7 | — | **13421.9** ± 26.5 | — |
  | `c0674bed` | flatten GraphEvaluator (custom gates) | **2979.8** ± 8.3 | **−7.3%** | **12478.6** ± 27.5 | **−7.0%** |
  | `5c3d837a` | + flatten lookup/trash | **2979.9** ± 16.6 | **−7.3%** | **12436.7** ± 20.3 | **−7.3%** |
  | `153a5a53` | + multi-element batching (B=4) | **2923.7** ± 16.4 | **−9.1%** | **12378.2** ± 25.2 | **−7.8%** |
  
CPU: Intel i5-11400H (6C/12T, 12 MB L3)
Turbo disabled for stability

The first 2 commits yield better improvements once the next PR, that introduces parallelism is added. Mainly, because parallelism increases memory pressure and the flattened evaluation gets more relevant.

### DIFF
--- a/proofs/src/plonk/evaluation.rs
+++ b/proofs/src/plonk/evaluation.rs
@@ -178,10 +178,12 @@ pub struct LookupGraphEvaluator<F: PrimeField> {
 }
 
 /// Evaluator
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct Evaluator<F: PrimeField> {
     ///  Custom gates evaluation
     pub custom_gates: GraphEvaluator<F>,
+    /// Flattened custom gates for fast evaluation.
+    pub custom_gates_flat: FlatGraphEvaluator<F>,
     ///  Lookups evaluation (one Vec per BatchedArgument, one entry per
     /// flattened arg)
     pub lookups: Vec<Vec<LookupGraphEvaluator<F>>>,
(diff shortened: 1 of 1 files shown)


proofs/src/plonk/evaluation.rs
```diff
@@ -230,10 +232,400 @@ pub struct CalculationInfo {
...
+        ) {
+            let mut check = |vs: &ValueSource| {
+                if matches!(
+                    vs,
+                    ValueSource::Fixed(..)
+                        | ValueSource::Advice(..)
+                        | ValueSource::Instance(..)
+                        | ValueSource::Challenge(..)
+                ) {
+                    if !reads.iter().any(|(v, _)| *v == *vs) {
```
miguel-ambrona: This is a linear search.

proofs/src/plonk/evaluation.rs
```diff
@@ -230,10 +232,400 @@ pub struct CalculationInfo {
...
+                ValueSource::TrashChallenge() => trash_challenge_idx,
+                ValueSource::Y() => y_idx,
+                ValueSource::PreviousValue() => previous_value_idx,
+                ValueSource::Fixed(..)
+                | ValueSource::Advice(..)
+                | ValueSource::Instance(..)
+                | ValueSource::Challenge(..) => {
+                    column_read_map
+                        .iter()
+                        .find(|(v, _)| *v == *vs)
```
miguel-ambrona: Linear search.

proofs/src/plonk/evaluation.rs
```diff
@@ -230,10 +232,400 @@ pub struct CalculationInfo {
...
+    pub a: u32,
+    pub b: u32,
+    pub c: u32,
+}
+
+/// Column read: load a value from a polynomial column at a rotated index.
+#[derive(Clone, Copy, Debug)]
+pub struct ColumnRead {
+    /// 0 = Fixed, 1 = Advice, 2 = Instance, 3 = Challenge.
+    pub col_type: u8,
```
miguel-ambrona: Why `u8`? Could we have an enum instead?

proofs/src/plonk/evaluation.rs
```diff
@@ -547,6 +551,21 @@ impl<F: PrimeField> GraphEvaluator<F> {
 }
 
 impl<F: PrimeField> FlatGraphEvaluator<F> {
+    /// Resolve a `ValueSource::Intermediate` handle to a flat buffer index.
+    /// Used to read specific outputs (e.g., lookup's sum_partial_products)
+    /// from the values buffer after evaluation.
+    pub fn resolve_idx(&self, vs: ValueSource) -> usize {
+        match vs {
+            ValueSource::Intermediate(idx) => {
+                // intermediates_offset = constants.len() + 5 + column_reads.len()
```
miguel-ambrona: Can we introduce a constant for this magic 5?

proofs/src/plonk/evaluation.rs
```diff
@@ -230,10 +232,400 @@ pub struct CalculationInfo {
...
+                    // In the flattened format, the column read is pre-loaded, so this
+                    // is just a copy: dst = src.
+                    let src = resolve(v);
+                    // Optimization: if dst == src, skip the copy. Otherwise emit Add with 0.
+                    if dst != src {
+                        ops.push(FlatOp {
+                            kind: FlatOpKind::Add,
+                            dst,
+                            a: src,
+                            b: 0, // Constant(0) = F::ZERO, add identity.
```
miguel-ambrona, line 523: This assumes that the first constant in the graph evaluator is 0, which is currently the case because `GraphEvaluator::default()` does so, but this is a bit fragile. How can we make it more robust?

proofs/src/plonk/evaluation.rs
```diff
@@ -398,33 +936,25 @@ impl<F: WithSmallOrderMulGroup<3>> Evaluator<F> {
...
+                flat.evaluate_chunk::<4, B>(
+                    &template,
+                    values,
+                    start,
+                    fixed,
+                    advice,
+                    instance,
+                    challenges,
+                    rot_scale.trailing_zeros(),
+                    (isize as u32).trailing_zeros(),
```
miguel-ambrona, line 1015: Shall we (debug) assert that `rot_scale` and `isize` are powers of 2? Is that an assumption that we are making?

davidnevadoc: It is, but a `debug_assert` does not hurt.
`rot_scale` must be the ratio between the extended and the original domain.
and `isize` is the domain size.

However, this is modified in the next PR (although I'm considering reverting the modification). Instead of passing the sizes, we pass the log sizes.

proofs/src/plonk/evaluation.rs
davidnevadoc: Good point. At first I thought this just went over the columns, so it was fine. But it goes over all value reads... Is that still fine?

In any case, the key point is that this will get called in key generation time, so it does not affect the prover performance. It is good to optimize key generation as well nonetheless

proofs/src/plonk/evaluation.rs
davidnevadoc: as above

proofs/src/plonk/evaluation.rs
davidnevadoc: Good point. Lets add a global cosntant named `NUM_CHALLENGES` to keep track of this.

proofs/src/plonk/evaluation.rs
miguel-ambrona: Good, let's forget about this then.

proofs/src/plonk/evaluation.rs
davidnevadoc: Right, it is quite ugly. Added a comment to keep track of this.

miguel-ambrona (approved):

PR #352, opened 2026-04-17
Prover Optimization: FFT

davidnevadoc:
Merges into #349 

 ## Summary

  1. **Cache twiddle factors**
Precompute the four twiddle arrays (forward/inverse × base/extended) once in `EvaluationDomain::new()`.
Extract `compute_twiddles` and `best_fft_with_twiddles` as public API.

  2. **Pruned DIF FFT**
 Fq-specialized Gentleman-Sande FFT exploiting the zero-padded structure of `coeff_to_extended` (n coefficients in a 4n  buffer). Skips all-zero subtrees and simplifies `(data, 0)` butterflies to a single multiply. 

## Performance Improvements

Performance improvement of this PR is not an accurate reflection of the final impact in the prover, since `opt_4` will increase FFT parallelization. It will be reported only in the meta PR.

## Future work
 - Integrate in blst to push performance improvements further: 
   * Coarse Montgomery reduction
   * Fused bit-reversal + stage 1
   * Prefetching
   * Improve Twiddle layout (contiguous access).
   * Inline ASM (no FFI boundary)
   * Flat loop + bitmask indexing
   * Thread model (Fixed slicing instead of  `rayon::join`)

### DIFF
--- a/curves/src/bls12_381/fq.rs
+++ b/curves/src/bls12_381/fq.rs
@@ -334,6 +334,14 @@ impl MulAssign<&Fq> for Fq {
     }
 }
 
+impl Fq {
+    /// Fused Gentleman-Sande (DIF) butterfly: `d = x0 - x1; x0 += x1; x1 = d * tw`.
+    #[inline]
+    pub fn gs_butterfly(x0: &mut Fq, x1: &mut Fq, twiddle: &Fq) {
+        unsafe { blst_fr_gs_bfly(&mut x0.0, &mut x1.0, &twiddle.0) };
+    }
+}
+
 impl<T> Sum<T> for Fq
 where
     T: Borrow<Fq>,
--- a/curves/src/bls12_381/fq.rs
+++ b/curves/src/bls12_381/fq.rs
@@ -335,7 +335,8 @@ impl MulAssign<&Fq> for Fq {
 }
 
 impl Fq {
-    /// Fused Gentleman-Sande (DIF) butterfly: `d = x0 - x1; x0 += x1; x1 = d * tw`.
+    /// Fused Gentleman-Sande (DIF) butterfly: `d = x0 - x1; x0 += x1; x1 = d *
+    /// tw`.
     #[inline]
     pub fn gs_butterfly(x0: &mut Fq, x1: &mut Fq, twiddle: &Fq) {
         unsafe { blst_fr_gs_bfly(&mut x0.0, &mut x1.0, &twiddle.0) };
--- a/curves/src/bls12_381/fq.rs
+++ b/curves/src/bls12_381/fq.rs
@@ -335,8 +335,8 @@ impl MulAssign<&Fq> for Fq {
 }
 
 impl Fq {
-    /// Fused Gentleman-Sande (DIF) butterfly: `d = x0 - x1; x0 += x1; x1 = d *
-    /// tw`.
+    /// Fused Gentleman-Sande (DIF) butterfly:
+    /// `d = x0 - x1; x0 += x1; x1 = d * tw`.
     #[inline]
     pub fn gs_butterfly(x0: &mut Fq, x1: &mut Fq, twiddle: &Fq) {
         unsafe { blst_fr_gs_bfly(&mut x0.0, &mut x1.0, &twiddle.0) };
--- a/curves/src/fft.rs
+++ b/curves/src/fft.rs
@@ -1,5 +1,6 @@
 use std::any::TypeId;
 
+use curve25519_dalek::Scalar;
 use ff::Field;
 use group::{GroupOpsOwned, ScalarMulOwned};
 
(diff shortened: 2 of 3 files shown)


curves/src/fft.rs
```diff
@@ -18,55 +21,67 @@ where
...
-/// by $n$.
+/// Reverse the low `log_n` bits of `n`.
+fn bitreverse(mut n: usize, log_n: u32) -> usize {
+    let mut r = 0;
+    for _ in 0..log_n {
+        r = (r << 1) | (n & 1);
+        n >>= 1;
+    }
+    r
+}
```
miguel-ambrona:
```suggestion
fn bitreverse(n: usize, log_n: u32) -> usize {
    n.reverse_bits() >> (usize::BITS - log_n)
}
```

If you are not applying this suggestion, at least document that your current implementation mutates `n`.

curves/src/fft.rs
```diff
@@ -18,55 +21,67 @@ where
...
+/// `2^log_n`.
+pub fn compute_twiddles<Scalar: Field>(omega: &Scalar, log_n: u32) -> Vec<Scalar> {
+    let half_n = 1usize << (log_n - 1);
+    (0..half_n)
+        .scan(Scalar::ONE, |w, _| {
+            let tw = *w;
+            *w *= omega;
+            Some(tw)
+        })
+        .collect()
```
miguel-ambrona: Would avoid the functional style improve performance?

```suggestion
    let mut twiddles = Vec::with_capacity(half_n);

    let mut w = Scalar::ONE;
    for _ in 0..half_n {
        twiddles.push(w);
        w *= omega;
    }

    twiddles
```

curves/src/fft.rs
```diff
@@ -18,55 +21,67 @@ where
...
 pub fn best_fft<Scalar: Field, G: FftGroup<Scalar>>(a: &mut [G], omega: Scalar, log_n: u32) {
-    fn bitreverse(mut n: usize, l: usize) -> usize {
-        let mut r = 0;
-        for _ in 0..l {
-            r = (r << 1) | (n & 1);
-            n >>= 1;
-        }
-        r
-    }
+    let twiddles = compute_twiddles(&omega, log_n);
```
miguel-ambrona, line 51: Can we reuse the twiddles among different FFTs? Is that what we are doing with `best_fft_with_twiddles`?

proofs/src/poly/domain.rs
```diff
@@ -5,11 +5,28 @@ use std::marker::PhantomData;
...
+///
+/// Twiddle factors are the powers of the root of unity used in the
+/// butterfly operations of the Cooley-Tukey FFT:
+///   `[1, ω, ω², ..., ω^(n/2 - 1)]`
+/// where ω is the n-th root of unity in the field.
+///
+/// Four sets are stored: forward and inverse for both the base domain (size
+/// 2^k) and the extended domain (size 2^extended_k).
+#[derive(Clone, Debug)]
+struct CachedTwiddles<F: Field> {
```
miguel-ambrona, line 23: Nice

proofs/src/poly/domain.rs
```diff
@@ -328,7 +362,7 @@ impl<F: WithSmallOrderMulGroup<3>> EvaluationDomain<F> {
         // domain.
         parallelize(&mut a.values, |h, mut index| {
             for h in h {
-                *h *= &self.t_evaluations[index % self.t_evaluations.len()];
+                *h *= &self.t_evaluations[index & (self.t_evaluations.len() - 1)];
```
miguel-ambrona: Hum, can you document this trick?

I understand you are assuming that `t_evaluations.len()` is a power of 2? If so, can we (debug) assert it?

curves/src/fft.rs
```diff
@@ -18,55 +21,67 @@ where
...
+    for _ in 0..log_n {
+        r = (r << 1) | (n & 1);
+        n >>= 1;
+    }
+    r
+}
+
+/// Precompute twiddle factors `[1, ω, ω², …, ω^(n/2 - 1)]` for an FFT of size
+/// `2^log_n`.
+pub fn compute_twiddles<Scalar: Field>(omega: &Scalar, log_n: u32) -> Vec<Scalar> {
```
miguel-ambrona, line 32:
```suggestion
/// Precompute twiddle factors `[1, ω, ω², …, ω^(n/2 - 1)]` for an FFT of size
/// `n = 2^log_n`.
pub fn compute_twiddles<Scalar: Field>(omega: &Scalar, log_n: u32) -> Vec<Scalar> {
```

We should also document that passing `log_n = 0` will result in a panic.

curves/src/fft.rs
```diff
@@ -86,11 +101,109 @@ pub fn best_fft<Scalar: Field, G: FftGroup<Scalar>>(a: &mut [G], omega: Scalar,
...
+        for i in 0..h {
+            gs(a, i, i + h, &tw[i * tc]);
+        }
+    }
+
+    // After the split, each half holds exactly `min(nz, h)` potentially
+    // non-zero entries, placed at the front.
+    let child_nz = nz.min(h);
+    let (left, right) = a.split_at_mut(h);
+    rayon::join(
```
miguel-ambrona, line 183: This `rayon::join` is called even on small subproblems, shall we have a size threshold?

curves/src/fft.rs
```diff
@@ -86,11 +101,109 @@ pub fn best_fft<Scalar: Field, G: FftGroup<Scalar>>(a: &mut [G], omega: Scalar,
...
+/// Uses pruned DIF (Gentleman-Sande) when `G` is the BLS12-381 scalar field
+/// [`Fq`]; falls back to standard DIT for other fields. The TypeId check is
+/// monomorphized to a constant — zero runtime cost.
+pub fn fft_coeff_to_extended<Scalar: Field, G: FftGroup<Scalar>>(
+    a: &mut [G],
+    twiddles: &[Scalar],
+    log_n: u32,
+    n_real: usize,
+) {
+    if TypeId::of::<G>() == TypeId::of::<Fq>() {
```
miguel-ambrona: Shall we add a type check on `Scalar` too?

miguel-ambrona (approved):

proofs/src/poly/domain.rs
davidnevadoc: Shouldn't be here... will remove it

curves/src/fft.rs
davidnevadoc: This is a tremendo melón that we don´t want to open now. 
Will address it as part of the future work when we fine tune the FFT at a lower level.

curves/src/fft.rs
davidnevadoc: Nice! done

curves/src/bls12_381/fq.rs
```diff
@@ -334,6 +334,15 @@ impl MulAssign<&Fq> for Fq {
     }
 }
 
+impl Fq {
+    /// Fused Gentleman-Sande (DIF) butterfly: `d = x0 - x1; x0 += x1; x1 = d *
+    /// tw`.
```
miguel-ambrona:
```suggestion
    /// Fused Gentleman-Sande (DIF) butterfly:
    /// `d = x0 - x1; x0 += x1; x1 = d * tw`.
```

curves/src/fft.rs
davidnevadoc: done!

PR #356, opened 2026-04-27
Fix verifier evals

### LINKED ISSUE #355: Verifier panics when a fixed column has multiple rotation queries
### Context & versions:

Bug present in `main` and not in `next`

### Steps to reproduce:
```
cargo run --example ethereum_signature --release
```
### Actual behavior:

```
>> Testing Ethereum signature verification
 - Deserialising public key...
 - Deserialising signatures...
 - Setting up vk...
 - Proof generation...
 - Proof verification...

thread 'main' panicked at /Users/queremendi/iohk/midnight/midnight-zk/proofs/src/plonk/verifier.rs:302:25:
insertion index (is 24) should be <= len (is 22)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Expected behavior:

Example expected to run correctly.

iquerejeta:
Fixes #355

### DIFF
--- a/proofs/src/plonk/verifier.rs
+++ b/proofs/src/plonk/verifier.rs
@@ -290,13 +290,18 @@ where
         .map(|_| -> Result<Vec<_>, _> { read_n(transcript, vk.cs.advice_queries.len()) })
         .collect::<Result<Vec<_>, _>>()?;
 
-    // Read (num_fixed_columns - num_simple_selectors) evals and from the transcript
-    // and fill up the "missing" places with 1 (the transcript doesn't contain evals
-    // corresponding to multiplicative, simple selectors)
-    let mut fixed_evals = read_n(
-        transcript,
-        vk.cs.num_fixed_columns() - vk.cs.num_simple_selectors(),
-    )?;
+    // Read one eval per non-simple-selector fixed query from the transcript,
+    // then fill the "missing" places with 1 (the transcript doesn't contain evals
+    // corresponding to multiplicative, simple selectors).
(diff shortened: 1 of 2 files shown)


proofs/src/plonk/verifier.rs
```diff
@@ -290,13 +290,18 @@ where
...
-    let mut fixed_evals = read_n(
-        transcript,
-        vk.cs.num_fixed_columns() - vk.cs.num_simple_selectors(),
-    )?;
+    // Read one eval per non-simple-selector fixed query from the transcript,
+    // then fill the "missing" places with 1 (the transcript doesn't contain evals
+    // corresponding to multiplicative, simple selectors).
+    // We count queries (not unique columns) because the same column can appear
+    // multiple times in fixed_queries at different rotation points.
+    let num_non_selector_fixed_queries = vk
```
miguel-ambrona: `num_fixed_queries_ignoring_simple_selectors`

miguel-ambrona (dismissed):

proofs/src/plonk/verifier.rs
remalue: I suggest to do this:

```
let mut fixed_evals = read_n(
  transcript,
  vk.cs.fixed_queries().len() - vk.cs.num_simple_selectors(),
)?;
```

We have `fixed queries = simple selector queries + non-simple selector queries + non-selector fixed queries`, so this change gives us all fixed queries except those corresp. to simple selectors. This also matches the prover side.

miguel-ambrona (approved):

PR #357, opened 2026-04-29
Deterministic prover

iquerejeta:
This PR cherry-picks the commit of version `1.1.0`. It also adds a new feature, `deterministic-prover` to ensure deterministic proof generation when using foreign field chip. 

The problem was that the `ZkStdLib` calls `new` in the `synthesis` function. The function `new` creates the foreign ECC chip (which contains the random point used for MSM). It does not make sense to pass in a `rng` in synthesis, so we now use a deterministic rng in the computation of the point `R` when the `deterministic-prover` feature is enabled.

### DIFF
--- a/circuits/src/ecc/foreign/weierstrass_chip.rs
+++ b/circuits/src/ecc/foreign/weierstrass_chip.rs
@@ -35,11 +35,11 @@ use midnight_proofs::{
 };
 use num_bigint::BigUint;
 use num_traits::One;
-use rand::rngs::OsRng;
+use rand::{CryptoRng, RngCore};
 #[cfg(any(test, feature = "testing"))]
 use {
     crate::testing_utils::Sampleable, crate::utils::util::FromScratch,
-    midnight_proofs::plonk::Instance, rand::RngCore,
+    midnight_proofs::plonk::Instance, rand::rngs::OsRng,
 };
 
 use super::gates::weierstrass::{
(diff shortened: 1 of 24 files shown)


circuits/src/ecc/foreign/weierstrass_chip.rs
```diff
@@ -1708,32 +1708,34 @@ where
...
             self.base_field_chip
                 .native_gadget
-                .assert_equal_to_fixed(layouter, &r.is_id, false)?;
+                .assert_equal_to_fixed(layouter, &p.is_id, false)?;
 
             p
         };
 
+        // Cache the random point from the chip. Note that `neg_alpha` is never used for
+        // the base randomness.
```
miguel-ambrona: What do you mean never used? This is dangerous, if we ever use `WS = 1` this will be a bug. I would create the correct `neg_alpha`.

Let me add a commit with a fix.

circuits/src/ecc/foreign/weierstrass_chip.rs
```diff
@@ -956,19 +980,30 @@ where
...
+    /// The RNG is used to sample a random point used for incomplete addition
+    /// Soundness does not rely on this point being random, but completeness
+    /// does.
     pub fn new(
         config: &ForeignWeierstrassEccConfig<C>,
         native_gadget: &N,
         scalar_field_chip: &S,
     ) -> Self {
+        #[cfg(feature = "deterministic-prover")]
+        let mut rng = ChaCha20Rng::seed_from_u64(0x9F3A7C2E);
```
miguel-ambrona: I would make this more obviously non-random, to highlight the fact that it is not important.
How about `0xDEADBEEF`? Although I feel inspired today:

```suggestion
        let mut rng = ChaCha20Rng::seed_from_u64(0x5EEDAB1E);
```

miguel-ambrona (dismissed):
LGTM!

miguel-ambrona (approved):

PR #365, opened 2026-04-30
Remove deterministic features

iquerejeta:
The use of a HashMap cannot guarantee determinism of proof generation, meaning that the feature did not achieve its intended behaviour.


remalue (approved):

davidnevadoc (approved):
LGTM!

PR #375, opened 2026-05-05
Remove multi-proof support

miguel-ambrona:
Remove multi-proof support; `create_proof` and `prepare` now operate on a single circuit and take `instances: &[&[F]]` instead of `&[&[&[F]]]`.


proofs/src/plonk/verifier.rs
```diff
@@ -366,55 +299,32 @@ where
...
-            |(
-                (
-                    (
-                        (((committed_instances, instance_evals), advice_commitments), advice_evals),
-                        permutation,
-                    ),
-                    lookups,
-                ),
-                trash,
-            )| {
```
iquerejeta, line 387: we are definitely not going to miss these indents

iquerejeta (approved):

proofs/src/plonk/verifier.rs
miguel-ambrona: I won't!

PR #379, opened 2026-05-13
Optimize permutation and logup commitments

### LINKED ISSUE #378: Optimize permutation and logup commitments
Permuation and Logup polynomils are not random, they have some structure.
We can take advantage of this structure to get more sparse MSMs when committing.

davidnevadoc:
## Summary
Change the base of permutation and logup polynomials before commiting to get a sparser representation before running the MSM. These commitments are a significant % of proving time ( need to get updated profiling), so improvements here have noticeable impact on overall proving time.

## Performance

To check performance, run
  1. `./bench_commits.sh compile`
  2. Disable turbo for more consistent measurements: `echo 1 | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo`
      (Your machine may need a slightly different command, ask claude)
  3. `./bench_commits.sh run`
 
This optimization may be a bit more impactful when turbo is disabled.

| Commit       | What's enabled                                       | bitcoin_sig (k=15)        | Δ vs base                | cred_full (k=17)          | Δ vs base                  |
|:-------------|:-----------------------------------------------------|:-------------------------:|:------------------------:|:-------------------------:|:--------------------------:|
| `08800448`   | bench harness only (baseline)                        | **2367.80 ms** (sd 4.2)   | —                        | **9184.59 ms** (sd 25.8)  | —                          |
|  `15d2938` | + Δ on `perm_z` + `multiplicities`                   | **2156.25 ms** (sd 14.3)  | **−211.6 ms / −8.93%**   | **8387.39 ms** (sd 24.5)  | **−797.2 ms / −8.68%**     |
| **`95472c2`** | **+ Δ² wired for `logup_aggregator`**              | **2047.35 ms** (sd 9.2)   | **−320.5 ms / −13.53%**  | **7779.43 ms** (sd 28.2)  | **−1405.2 ms / −15.30%**   |

## TODO
 - [x] Update changelogs
 - [x] Drop benchmark commits


proofs/src/poly/kzg/params.rs
```diff
@@ -22,10 +22,31 @@ use crate::{
...
 
+/// Builds the suffix-sum of an affine SRS: `out[i] = sum_{j=i}^{n-1} input[j]`.
+///
+/// Accumulate in projective coordinates (one curve add per step) and batch-normalise
+/// to affine at the end. n-1 projective adds + one O(n) batch_normalise.
+fn suffix_sum<C: CurveAffine>(input: &[C]) -> Vec<C> {
+    let n = input.len();
+    if n == 0 {
+        return Vec::new();
+    }
```
miguel-ambrona: Will we ever give an empty input? I guess this is fine, probably better than panicking, but I guess we won't use it.

proofs/src/poly/mod.rs
```diff
@@ -52,6 +52,8 @@ pub enum PolynomialBasis {
     Lagrange,
     /// Lagrange basis over an extended coset domain.
     ExtendedLagrange,
+    /// Contiguous friendly = Lagrange where b_i := a_i - a_(i-1)
```
miguel-ambrona: This comment could be revisited. Please, add more details.

proofs/src/poly/mod.rs
```diff
@@ -184,6 +186,43 @@ impl PolynomialRepresentation for ExtendedLagrangeCoeff {
...
+    fn k<F: WithSmallOrderMulGroup<3>>(evaluation_domain: &EvaluationDomain<F>) -> u32 {
+        evaluation_domain.k()
+    }
+
+    fn coeff_to_self<F: WithSmallOrderMulGroup<3>>(
+        _evaluation_domain: &EvaluationDomain<F>,
+        _poly: Polynomial<F, Coeff>,
+    ) -> Polynomial<F, Self> {
+        unimplemented!("LagrangeDelta is constructed from LagrangeCoeff via `into_delta`.")
+    }
```
miguel-ambrona, line 233: Well, you could convert to Lagrange and then into_delta. But I guess failing is better to force us think of the actual conversions where needed.

proofs/src/poly/mod.rs
```diff
@@ -184,6 +186,43 @@ impl PolynomialRepresentation for ExtendedLagrangeCoeff {
...
+
+    fn g_coset<F: WithSmallOrderMulGroup<3>>(_evaluation_domain: &EvaluationDomain<F>) -> F {
+        F::ONE
+    }
```
miguel-ambrona, line 278: Is this right? Should this be unimplemented too?

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
     pub(crate) _marker: PhantomData<B>,
 }
 
+impl<F: Field> Polynomial<F, LagrangeCoeff> {
+    /// Convert a Lagrange-form polynomial into its delta encoding,
+    /// `b_0 = a_0`, `b_i = a_i - a_{i-1}`, in place.
```
miguel-ambrona: specify, for all i > 0?

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
...
+    pub fn into_delta(mut self) -> Polynomial<F, LagrangeDeltaCoeff> {
```
miguel-ambrona: This mutates self, shall the doc mention it or not since it is implied by the fn signature?

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
...
+        for i in (1..self.values.len()).rev() {
+            let prev = self.values[i - 1];
+            self.values[i] -= prev;
```
miguel-ambrona: Is the auxiliary variable needed?

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
...
+        }
+        Polynomial {
+            values: self.values,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Borrowing variant of [`Self::into_delta`]: produce the delta
+    /// encoding in a freshly allocated buffer, leaving `self` untouched.
```
miguel-ambrona: Is `into_` vs `to_` the standard terminology for such difference?

Would `_inplace` make sense?

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
...
+            out.push(self.values[0]);
+        }
+        for i in 1..n {
+            out.push(self.values[i] - self.values[i - 1]);
+        }
+        Polynomial {
+            values: out,
+            _marker: PhantomData,
+        }
+    }
```
miguel-ambrona, line 336: Is this more efficient than cloning and then calling the inplace version?

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
...
+}
+
+impl<F: Field> Polynomial<F, LagrangeDeltaCoeff> {
+    /// Inverse of `into_delta`: prefix-sum the deltas back to Lagrange form,
+    /// `a_0 = b_0`, `a_i = a_{i-1} + b_i`, in place.
+    pub fn into_lagrange(mut self) -> Polynomial<F, LagrangeCoeff> {
+        for i in 1..self.values.len() {
+            let prev = self.values[i - 1];
+            self.values[i] += prev;
```
miguel-ambrona, line 348: Is the aux variable `prev` necessary?

proofs/src/poly/mod.rs
miguel-ambrona: Also, if the function is inplace, why bother returning a Polynomial? This could return unit.

bench_commits.sh
```diff
@@ -35,9 +35,9 @@ REPO_ROOT=$(git rev-parse --show-toplevel)
 # entries in zk_stdlib/Cargo.toml) but no perf-relevant change. Each
 # subsequent entry should be a successive perf-relevant commit.
 COMMITS=(
-    # <baseline-sha>   # bench infrastructure on top of main, no perf change.
-    # <delta-sha>      # LagrangeDelta wired for perm_z + multiplicities.
-    # <ddelta-sha>     # LagrangeDoubleDelta wired for logup_aggregator.
+ 08800448 # bench infrastructure on top of main, no perf change.
+ 244952e7 # LagrangeDelta wired for perm_z + multiplicities.
+ 1c364bb8 # LagrangeDoubleDelta wired for logup_aggregator.
 )
```
miguel-ambrona: Do we wanna merge this?

miguel-ambrona (dismissed):
Amazing! It is impressive that you keep coming up with such massive optimizations.

I leave some minor comments, most of them (on `delta`) apply to the `double_delta`.

bench_commits.sh
davidnevadoc: No, this will get dropped, together with the benchmarks

proofs/src/poly/mod.rs
davidnevadoc: I don't know. The difference will be minimal anyways so I think we should choose the clearer option.
Cloning + in-place is quite clean, but it may be worse from the prover perspective... 🤔
On the other hand, if in the future the prover does not need to use the lagrange version further down the protocol, he may easily drop the clone. Do we want to give the prover code the power / responsibility to handle memory in this sense?

proofs/src/poly/mod.rs
davidnevadoc: To the first comment, it is implicit because it takes `mut self` and because of the `into_` prefix. To the second, is convenient to return it because this way we can do things like 
`commit( poly.into_delta())`.

proofs/src/poly/mod.rs
davidnevadoc: No, but honestly I can't fully understand why. The error is:
`cannot borrow `self.values` as immutable because it is also borrowed as mutable`

proofs/src/poly/mod.rs
davidnevadoc: I think it is. I believe `into_` means the input will get consumed. I am not sure if `to_` carries some strict implications.

proofs/src/poly/mod.rs
davidnevadoc: Yes, see comment above

proofs/src/poly/kzg/params.rs
davidnevadoc: No, but I don't think this hurts performance and the alternative is panicking so I think it is fine

proofs/src/poly/mod.rs
davidnevadoc: Done

proofs/src/poly/mod.rs
davidnevadoc: Yes. I don't like this trait very much, in some aspects I find it is not general or expressive enough, and in others it is weirdly specific.

proofs/src/poly/mod.rs
davidnevadoc: Removed. It is not necessary so better to leave as unimplemented.

proofs/src/poly/mod.rs
davidnevadoc: Done

proofs/src/poly/kzg/params.rs
davidnevadoc: In the end I removed it, since this function should never receive an empty input.

proofs/src/poly/mod.rs
davidnevadoc: I opened an issue to fix this #382.
It may be good to wait for fflonk, in case it needs more changes on how we split and commit to polynomials.

proofs/src/poly/kzg/params.rs
miguel-ambrona: But did you document the panic?

proofs/src/poly/mod.rs
miguel-ambrona: I would clone + inplace.

proofs/src/poly/mod.rs
```diff
@@ -221,11 +229,11 @@ impl PolynomialRepresentation for LagrangeDeltaCoeff {
         _evaluation_domain: &EvaluationDomain<F>,
         _poly: Polynomial<F, Coeff>,
     ) -> Polynomial<F, Self> {
-        unimplemented!("LagrangeDelta is constructed from LagrangeCoeff via `into_delta`.")
+        unimplemented!("LagrangeDelta is constructed from LagrangeCoeff via `to_delta`.")
     }
 
     fn g_coset<F: WithSmallOrderMulGroup<3>>(_evaluation_domain: &EvaluationDomain<F>) -> F {
-        F::ONE
+        unimplemented!()
```
miguel-ambrona: Message? Not necessary though.

miguel-ambrona (dismissed):

proofs/src/poly/mod.rs
davidnevadoc: I decided to go for the borrowing version. Turns out simple allocation is a bit cheaper than cloning. And this way the code in the prover is a bit cleaner.

proofs/src/poly/kzg/params.rs
davidnevadoc: Done now

proofs/src/poly/mod.rs
davidnevadoc: Function removed in favour of the borrowing version.

proofs/src/poly/mod.rs
```diff
@@ -192,6 +231,58 @@ pub struct Polynomial<F, B> {
...
+        if n >= 1 {
+            out.push(self.values[0]);
+        }
+        for i in 1..n {
+            out.push(self.values[i] - self.values[i - 1]);
+        }
+        Polynomial {
+            values: out,
+            _marker: PhantomData,
+        }
```
iquerejeta, line 335: What about 

```suggestion
        let mut out = self.clone();
        out.into_delta()
```

proofs/src/poly/mod.rs
```diff
@@ -266,6 +314,58 @@ impl<F: Field> Polynomial<F, LagrangeCoeff> {
...
+        if n >= 2 {
+            out.push(self.values[1] - self.values[0].double());
+        }
+        for i in 2..n {
+            out.push(self.values[i] - self.values[i - 1].double() + self.values[i - 2]);
+        }
+        Polynomial {
+            values: out,
+            _marker: PhantomData,
+        }
```
iquerejeta, line 335: Ditto, can you do
```suggestion
        let mut out = self.clone();
        out.into_double_delta();
```

iquerejeta (commented):
What is the performance of running the suffix sum? 
```
        let g_lagrange_delta = suffix_sum(&g_lagrange);
        let g_lagrange_double_delta = suffix_sum(&g_lagrange_delta);
```

While it might be cheap, we are adding that to an operation that is essentially free `(read_pk)`. I don't think it'll be an issue, but I'd like to understand what is the cost of it.

davidnevadoc:
> What is the performance of running the suffix sum?
> 
> ```
>         let g_lagrange_delta = suffix_sum(&g_lagrange);
>         let g_lagrange_double_delta = suffix_sum(&g_lagrange_delta);
> ```
> 
> While it might be cheap, we are adding that to an operation that is essentially free `(read_pk)`. I don't think it'll be an issue, but I'd like to understand what is the cost of it.

It's not expensive, but it is significant considering that the rest of `read_pk` is basically free.
I don't have exact numbers but for `k<=18` it could take up to 0.5s.
There are easy ways to mitigate this:
 - Parallelize the prefix sum with a Blelloch scan ( we already did this with good results in the grand product computation)
 - Store them in the SRS,  just as the Lagrange basis
 
 We won't address this issue in this PR, as it needs a some further profiling and discussion. Tracked in #388

proofs/src/poly/mod.rs
davidnevadoc: The inplace versions (`into_`) where removed in favour of these, since the original poly is always needed downstream in the protocol.

proofs/src/poly/mod.rs
davidnevadoc: Same as above

proofs/src/poly/mod.rs
davidnevadoc: The inplace versions (`into_`) where removed in favour of these, since the original poly is always needed downstream in the protocol.

proofs/src/poly/mod.rs
davidnevadoc: Same as above

miguel-ambrona (approved):

PR #381, opened 2026-05-18
Refactor commitment labels and commitment references

miguel-ambrona:
Move commitment labels to actual commitments (and not queries).

Move `OnePiece`/`Linear` abstraction to KZG commitments (through a new `KZGCommitment` enum), simplifying the type of `CommitmentReference`.

This allows us to truly use the homomorphic requirements in the PCS which were not being used before (the homomorphic property of KZG commitments was hard-coded in the multi-open argument).

### DIFF
--- a/proofs/src/plonk/linearization/verifier.rs
+++ b/proofs/src/plonk/linearization/verifier.rs
@@ -2,14 +2,12 @@ use std::collections::BTreeMap;
 
 use ff::PrimeField;
 
-use crate::{
-    plonk::VerifyingKey,
-    poly::{commitment::PolynomialCommitmentScheme, CommitmentLabel, VerifierQuery},
-};
+use crate::{plonk::VerifyingKey, poly::commitment::PolynomialCommitmentScheme};
 
-/// Construct the commitment to the linearization polynomial
-/// (which will be checked that it opens to `0` at `x` in the multi-open
-/// argument):
+/// Construct the commitment to the linearization polynomial and its expected
+/// evaluation at `x`.
+///
+/// The commitment is:
 ///
 ///  `S_0 * id_0(x) + y * S_1 * id_1(x) + ... + y^m * S_m * id_m(x)
 ///        - (h_0 + x^{n-1} * h_1 + ... + x^{l*(n-1)} * h_l) * (x^n-1),`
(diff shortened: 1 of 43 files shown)


proofs/src/poly/kzg/commitment.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+impl<E: MultiMillerLoop> KZGCommitment<E> {
+    /// Extracts the inner curve point, panicking if this is a `Linear`
+    /// commitment.
+    pub fn into_point(self) -> E::G1 {
+        match self {
+            Self::Simple(p, _) => p,
+            Self::Linear(..) => panic!("expected KZGCommitment::Simple"),
```
iquerejeta, line 62: What about evaluating the actual point?

proofs/src/poly/kzg/commitment.rs
```diff
@@ -0,0 +1,161 @@
...
+};
+
+/// A KZG commitment: either a single curve point, or a lazy linear combination
+/// of curve points kept unevaluated for MSM batching.
+///
+/// Each point carries a [`CommitmentLabel`] that identifies its role in the
+/// protocol (e.g. `Fixed(i)`, `Advice(i)`).  Labels are protocol-level
+/// metadata; they are not included in serialization or hashing.
+#[derive(Clone, Debug)]
+pub enum KZGCommitment<E: MultiMillerLoop> {
```
iquerejeta, line 33: Feels like a few natural operations are missing: 
* Convert a linear representation into simple representations (e.g. to be able to commit to it)
* Creating a `Linear` commitment out of simple commitments (you even mention below that that might happen)
* Update the label of a KZGCommitment. Ser/Deser looses the label info, so its likely that a verifier will want to label a commitment once created.

proofs/src/poly/kzg/commitment.rs
```diff
@@ -0,0 +1,161 @@
...
+use midnight_curves::pairing::MultiMillerLoop;
+
+use crate::{
+    poly::query::CommitmentLabel,
+    transcript::{Hashable, TranscriptHash},
+    utils::helpers::{ProcessedSerdeObject, SerdeFormat},
+};
+
+/// A KZG commitment: either a single curve point, or a lazy linear combination
+/// of curve points kept unevaluated for MSM batching.
```
iquerejeta, line 16: Can we extend the documentation here? It should be clear why we are representing commitments in this way.

proofs/src/poly/kzg/commitment.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    fn add(self, other: Self) -> Self {
+        let (mut points, mut scalars, mut labels) = match self {
+            Self::Simple(p, label) => (vec![p], vec![E::Fr::ONE], vec![label]),
+            Self::Linear(points, scalars, labels) => (points, scalars, labels),
+        };
+        let (other_points, other_scalars, other_labels) = match other {
+            Self::Simple(p, label) => (vec![p], vec![E::Fr::ONE], vec![label]),
+            Self::Linear(points, scalars, labels) => (points, scalars, labels),
+        };
```
iquerejeta, line 168: It would be more efficient to match (self, other) and if both are linear, you just do the addition. Otherwise, you might include it in a larger msm, and treat the ones as if they were full length scalars.

proofs/src/poly/kzg/mod.rs
```diff
@@ -273,25 +273,9 @@ where
     ) -> Result<DualMSM<E>, Error>
     where
         E::Fr: Sampleable<T::Hash> + Ord + Hash + Hashable<T::Hash>,
-        E::G1: 'com + CurveExt<ScalarExt = E::Fr>,
-        KZGCommitment<E>: Hashable<T::Hash>,
+        E::G1: CurveExt<ScalarExt = E::Fr>,
+        KZGCommitment<E>: Hashable<T::Hash> + 'com,
     {
-        // Add dummy queries to reduce the number of distinct multi-open point sets.
-        #[cfg(feature = "fewer-point-sets")]
```
iquerejeta, line 255: removing this feature?

proofs/src/plonk/linearization/verifier.rs
```diff
@@ -18,58 +16,35 @@ use crate::{
...
 /// # Returns
 ///
-/// A [VerifierQuery], that checks if the commitment to the linearization
-/// polynomial opens to `0` at the evaluation challenge `x`. The commitment
-/// itself is an MSM represented as a vector of points and a vector
-/// of scalars.
-#[allow(clippy::type_complexity)]
+/// `(commitment, expected_eval)` where the commitment to the linearization
+/// polynomial is expected to open to `expected_eval` at `x`.
 #[allow(clippy::too_many_arguments)]
```
iquerejeta: We may be able to remove this flag now (I think the boundary is at 7)

proofs/src/poly/query.rs
```diff
@@ -96,70 +96,28 @@ impl<'com, F: PrimeField> Query<F> for ProverQuery<'com, F> {
...
-///     * commitments (representing, either, commitments to simple,
-///       multiplicative selectors or the commitment to the constant polynomial
-///       `P(X) = 1`).
-/// The linear combination is given in form of two vectors: one holds references
-/// to the commitments, the other one holds the scalars.
-pub enum CommitmentReference<'com, F: PrimeField, CS: PolynomialCommitmentScheme<F>> {
-    OnePiece(&'com CS::Commitment),
-    Linear(Vec<&'com CS::Commitment>, Vec<F>, Vec<CommitmentLabel>),
+/// Two `CommitmentReference`s are equal iff they point to the same allocation,
+/// so that commitments are grouped by identity rather than by value.
```
iquerejeta:
```suggestion
/// so that commitments are grouped by reference rather than by value.
```

iquerejeta (commented):
Minor comments on the associated functions of KZG commitments.

proofs/src/poly/kzg/commitment.rs
miguel-ambrona: What do you mean?

proofs/src/poly/kzg/commitment.rs
miguel-ambrona: I'm not sure what you mean by (1).

(2) is possible through the homomorphic property.

(3) That is intentional. We do not wanna waste space in the proof with labels. The verifier knows exactly what they are reading and how to label it.

proofs/src/poly/kzg/commitment.rs
miguel-ambrona: Isn't the lazy linear combination comment clear?

proofs/src/poly/kzg/commitment.rs
miguel-ambrona: What do you suggest exactly? I doubt it can be much cleaner than this.

proofs/src/poly/kzg/mod.rs
miguel-ambrona: I should not have committed that, thanks.

iquerejeta (approved):

PR #383, opened 2026-05-19
Implement various traits for `AssignedScalar` of the native ECC chip

miguel-ambrona:
These are implemented through the BigUint gadget (instead of foreign field emulation) so that there is no overhead in the architecture of circuits using Jubjub.

These new traits are necessary for ZKIRv3 after we agreed on that Compact will allow arithmetic operations on Jubjub scalars.

### DIFF
--- a/circuits/src/ecc/native/edwards_chip.rs
+++ b/circuits/src/ecc/native/edwards_chip.rs
@@ -198,7 +198,7 @@ impl<C: CircuitCurve> AssignedScalarOfNativeCurve<C> {
 
     /// Constructs an [`AssignedScalarOfNativeCurve`] from an
     /// [`AssignedBigUint`]. The result is not guaranteed to be canonical but is
-    /// guaranteed to have at most [`C::ScalarField::NUM_BITS`] bits.
+    /// guaranteed to have at most `C::ScalarField::NUM_BITS` bits.
     fn from_biguint(
         layouter: &mut impl Layouter<C::Base>,
         biguint_gadget: &BigUintGadget<C::Base, NG<C::Base>>,
--- a/circuits/src/ecc/native/edwards_chip.rs
+++ b/circuits/src/ecc/native/edwards_chip.rs
@@ -171,32 +171,45 @@ fn reduce_biguint_mod_scalar_order<C: CircuitCurve>(
 }
 
 impl<C: CircuitCurve> AssignedScalarOfNativeCurve<C> {
-    /// Converts the given scalar to a BigUint reduced modulo the scalar field
-    /// order.
-    fn to_reduced_biguint(
+    /// Converts the scalar's bits into an [`AssignedBigUint`].
+    /// The result is not guaranteed to be in canonical form (i.e. it may
+    /// represent a value greater than or equal to the scalar field order).
+    fn into_biguint(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        biguint_gadget: &BigUintGadget<C::Base, NG<C::Base>>,
+    ) -> Result<AssignedBigUint<C::Base>, Error> {
+        biguint_gadget.from_le_bits(layouter, &self.bits)
+    }
+
+    /// Converts the scalar's bits into an [`AssignedBigUint`].
+    /// The result is guaranteed to be in canonical form.
+    fn into_canonical_biguint(
         &self,
         layouter: &mut impl Layouter<C::Base>,
         biguint_gadget: &BigUintGadget<C::Base, NG<C::Base>>,
     ) -> Result<AssignedBigUint<C::Base>, Error> {
         let s = biguint_gadget.from_le_bits(layouter, &self.bits)?;
-        if self.bits.len() < C::ScalarField::NUM_BITS as usize {
-            return Ok(s);
-        }
         reduce_biguint_mod_scalar_order::<C>(layouter, biguint_gadget, &s)
     }
 
     /// Constructs an [`AssignedScalarOfNativeCurve`] from an
-    /// [`AssignedBigUint`], reducing it modulo the scalar field order if
-    /// necessary to ensure the resulting bit representation is canonical.
+    /// [`AssignedBigUint`]. The result is not guaranteed to be canonical but is
+    /// guaranteed to have at most [`C::ScalarField::NUM_BITS`] bits.
     fn from_biguint(
         layouter: &mut impl Layouter<C::Base>,
         biguint_gadget: &BigUintGadget<C::Base, NG<C::Base>>,
         s: &AssignedBigUint<C::Base>,
     ) -> Result<Self, Error> {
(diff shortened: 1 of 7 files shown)


circuits/src/ecc/native/edwards_chip.rs
```diff
@@ -556,7 +603,12 @@ impl<C: EdwardsCurve> EccChip<C> {
         if point.checked_in_subgroup {
             return Err(Error::Synthesis("clear_cofactor() should not be called in a point that is already guaranteed to be in the prime-order subgroup.".to_owned()));
         }
-        let r = self.mul_by_constant(layouter, C::ScalarField::from_u128(C::COFACTOR), point)?;
+        let r = EccInstructions::mul_by_constant(
+            self,
+            layouter,
+            C::ScalarField::from_u128(C::COFACTOR),
+            point,
+        )?;
```
iquerejeta, line 639: This is odd - why are you making these changes? i.e., why not `self.mul_by_constant`?

circuits/src/ecc/native/edwards_chip.rs
```diff
@@ -952,8 +1051,146 @@ impl<C: EdwardsCurve> EqualityInstructions<C::Base, AssignedNativePoint<C>> for
...
+    }
+
+    fn inv(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        s: &AssignedScalarOfNativeCurve<C>,
+    ) -> Result<AssignedScalarOfNativeCurve<C>, Error> {
+        let one = self.assign_fixed(layouter, C::ScalarField::ONE)?;
+        self.div(layouter, &one, s)
+    }
```
iquerejeta, line 1216: Makes more sense to do it the other way around. Implement `inv` and use it in `div`. Otherwise you are doing some unecessary ops, like: 

```
let r = r.to_reduced_biguint(layouter, &self.biguint_gadget)?;
```

for fixed 1.

circuits/src/ecc/native/edwards_chip.rs
```diff
@@ -952,8 +1080,156 @@ impl<C: EdwardsCurve> EqualityInstructions<C::Base, AssignedNativePoint<C>> for
...
+        s.value().error_if_known_and(|s| s.is_zero_vartime())?;
+
+        let res_val = r.value().zip(s.value()).map(|(r, s)| r * s.invert().unwrap());
+        let res = self.biguint_gadget.assign_biguint(
+            layouter,
+            res_val.map(|a| a.to_biguint()),
+            C::ScalarField::NUM_BITS,
+        )?;
+
+        let r = r.to_biguint(layouter, &self.biguint_gadget)?;
```
iquerejeta, line 1198: For the equality check
```
        let r = r.to_canonical_biguint(layouter, &self.biguint_gadget)?;
```

circuits/src/ecc/native/edwards_chip.rs
```diff
@@ -144,6 +152,68 @@ impl<C: EdwardsCurve> Sampleable for AssignedScalarOfNativeCurve<C> {
...
+
+    /// Converts the scalar's bits into an [`AssignedBigUint`].
+    /// The result is guaranteed to be in canonical form.
+    fn to_canonical_biguint(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        biguint_gadget: &BigUintGadget<C::Base, NG<C::Base>>,
+    ) -> Result<AssignedBigUint<C::Base>, Error> {
+        let s = biguint_gadget.from_le_bits(layouter, &self.bits)?;
+        reduce_biguint_mod_scalar_order::<C>(layouter, biguint_gadget, &s)
```
iquerejeta, line 196: Shouldn't we do the reduction only if we are not enforcing this to be canonical?

circuits/src/ecc/native/edwards_chip.rs
miguel-ambrona: Because `mul_by_constant` exists both in `ArithInstructions` and `EccInstructions`.

Rust could figure out which from the types but does not.

circuits/src/ecc/native/edwards_chip.rs
miguel-ambrona: I believe that is not that bad after the optimizations in the next commits.

Implementing `div` from `inv` would also be more expensive than the current `div`.

Anyway, let me implement `inv` independently and see.

circuits/src/ecc/native/edwards_chip.rs
miguel-ambrona: What?

circuits/src/ecc/native/edwards_chip.rs
miguel-ambrona: Not needed as the honest prover will use the canonical `r`.

Remember that we allow ourselves to have potentially incomplete equality assertions as long as the honest prover is complete in those.

circuits/src/ecc/native/edwards_chip.rs
miguel-ambrona: Ok, I implemented `inv` directly as:

```rust
    fn inv(
        &self,
        layouter: &mut impl Layouter<C::Base>,
        s: &AssignedScalarOfNativeCurve<C>,
    ) -> Result<AssignedScalarOfNativeCurve<C>, Error> {
        s.value().error_if_known_and(|s| s.is_zero_vartime())?;

        let res_val = s.value().map(|s| s.invert().unwrap());
        let res = self.biguint_gadget.assign_biguint(
            layouter,
            res_val.map(|a| a.to_biguint()),
            C::ScalarField::NUM_BITS,
        )?;

        let s = s.to_biguint(layouter, &self.biguint_gadget)?;
        let res_times_s = self.biguint_gadget.mul(layouter, &res, &s)?;
        let reduced_res_times_s =
            reduce_biguint_mod_scalar_order::<C>(layouter, &self.biguint_gadget, &res_times_s)?;

        self.biguint_gadget.assert_equal_to_fixed(
            layouter,
            &reduced_res_times_s,
            BigUint::from(1u64),
        )?;

        AssignedScalarOfNativeCurve::from_biguint(layouter, &self.biguint_gadget, &res)
    }
```

And there is literally no change in the goldenfiles.

I'll leave it as is.

circuits/src/ecc/native/edwards_chip.rs
iquerejeta: We have the field `enforced_canonical`. Shouldn't we check that before doing the reduction?

miguel-ambrona: Oh yeah, that was the whole point. I had the impression I was missing something yesterday. Thanks!

miguel-ambrona: Done, this was a big improvement.

iquerejeta (dismissed):

iquerejeta (approved):

erik-axeljakobsson (approved):

PR #392, opened 2026-05-22
Label all polynomials

miguel-ambrona:
A small refactor preparing the ground for the Multi PCS interface.

We label all polynomials (all commitments since each commitment holds a single polynomial for now). Labels are not serialized. After a commitment is deserialized, the verifier has to explicitly call `fn label` to include the label they expect for it.


proofs/src/plonk/circuit.rs
```diff
@@ -1749,32 +1749,12 @@ impl<F: Field> ConstraintSystem<F> {
...
-            table_map,
-        ));
-
-        index
+        self.batch_lookup_any(name, selector, |cells| {
+            table_map(cells)
+                .into_iter()
+                .map(|(inputs, table)| (inputs, cells.query_fixed(table.inner(), Rotation::cur())))
+                .collect()
+        })
```
irakoton, line 1757: Is this something that has been forgotten in a previous PR? Batch lookups seem to be unrelated to the change

proofs/src/plonk/circuit.rs
```diff
@@ -1806,13 +1806,14 @@ impl<F: Field> ConstraintSystem<F> {
...
-        self.lookups.push(logup::BatchedArgument::new(
-            name.as_ref(),
-            selector,
-            table_map,
-        ));
+        // Append the ordinal to guarantee uniqueness even when the caller supplies
+        // the same name twice (e.g. the same chip configured more than once).
+        let name = format!("{} lookup: {}", ordinal(index + 1), name.as_ref());
+
+        self.lookups.push(logup::BatchedArgument::new(&name, selector, table_map));
```
irakoton, line 1816: ~IIUC, you use `ordinal` to ensure that several consecutive configurations of the same chip get unique name, but we still have to enforce manually to give different names to each gate call. Is this requirement documented somewhere?~

Nevermind, I misread the `format!`. Everything is unique by construction.

proofs/src/utils/helpers.rs
```diff
@@ -111,6 +111,18 @@ where
...
+    let suffix = match (n % 100, n % 10) {
+        (11..=13, _) => "th",
+        (_, 1) => "st",
+        (_, 2) => "nd",
+        (_, 3) => "rd",
+        _ => "th",
+    };
+    format!("{n}{suffix}")
+}
+
```
irakoton: We could also just do `lookup #n` instead of using ordinals, I think it's as readable. No strong opinion.

proofs/src/poly/commitment.rs
```diff
@@ -72,6 +73,16 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
+/// A commitment that can be assigned a [`PolynomialLabel`].
+///
+/// Deserialized commitments start with [`PolynomialLabel::NoLabel`]; call sites
+/// must attach the correct label before the commitment reaches the MSM layer.
+pub trait Labelable {
+    /// Attaches the given [`PolynomialLabel`] to this commitment, replacing any
+    /// existing label (including [`PolynomialLabel::NoLabel`]).
+    fn label(self, label: PolynomialLabel) -> Self;
+}
+
```
irakoton, line 85: I don't really like this part of the API, but I thought about it for a while, and I don't think we can do better. The only reasonable other choice would be to also serialise the label in the transcript, but apart from the proof-size overhead (played with Claude and got a +5% estimate), I'm not comfortable with adding yet another possible way of doing Fiat-Shamir incorrectly.

Is the plan to stick to this API unless someone has a genius improvement idea?

proofs/src/poly/query.rs
```diff
@@ -10,29 +10,57 @@ use crate::{
...
+            Self::PermutationFixed(i) => write!(f, "perm_fixed_{i}"),
+            Self::PermutationAccumulator(i) => write!(f, "perm_acc_{i}"),
+            Self::LogupHelper(name) => write!(f, "logup_helper({name})"),
+            Self::LogupMultiplicities(name) => write!(f, "logup_multiplicities({name})"),
+            Self::LogupAggregator(name) => write!(f, "logup_aggregator({name})"),
+            Self::Trash(name) => write!(f, "trash({name})"),
+            Self::Quotient => f.write_str("quotient"),
+            Self::QuotientPiece(i) => write!(f, "quotient_piece_{i}"),
+            Self::Collapsed => f.write_str("collapsed"),
             Self::Custom(s) => f.write_str(s),
```
irakoton: Kind of a improbable degenerate side case, but maybe change the display to avoid cases where someone uses another label name as Custom (e.g., `collapsed`). Maybe `custom({s})`.

proofs/src/poly/query.rs
```diff
@@ -10,29 +10,57 @@ use crate::{
...
+    /// PLONK quotient polynomial h(X), committed as a single piece.
+    Quotient,
+    /// PLONK quotient polynomial h(X), committed in pieces (piece index).
+    QuotientPiece(usize),
+    /// Trash compressed polynomial, for a named argument.
+    Trash(String),
+    /// A commitment obtained by collapsing an MSM.
+    Collapsed,
     /// User-defined label.
     Custom(String),
```
irakoton, line 40: IIUC, the `Custom` variant will only be used in the implems of multi-open/multi-prepare? Not sure where else we would need something custom. In fflonk, we will likely use it there as well.

circuits/src/verifier/accumulator.rs
```diff
@@ -94,26 +94,29 @@ impl<S: SelfEmulation> Accumulator<S> {
...
+                    PolynomialLabel::PermutationFixed(i) => {
                         let name = perm_commitment_name(prefix, *i);
                         assert_eq!(fixed_bases.get(&name), Some(base));
                         fixed_base_scalars.insert(name, *scalar);
                     }
-                    CommitmentLabel::Custom(s) if s == "-G" => {
+                    PolynomialLabel::Custom(s) if s == "-G" => {
                         assert_eq!(fixed_bases.get(s), Some(base));
                         fixed_base_scalars.insert("-G".into(), *scalar);
                     }
```
irakoton, line 116: I'm realising it now, but does this mean that whenever a new `Custom` is used in a new implementation (typically fflonk), the code of the verifier circuit has to be adapted to take the new variants into account?

circuits/src/verifier/mod.rs
```diff
@@ -47,6 +47,7 @@ pub use accumulator::{Accumulator, AssignedAccumulator};
 #[derive(Clone, Debug)]
 pub(crate) struct LabeledPoint<S: SelfEmulation> {
     pub(crate) point: S::AssignedPoint,
+    #[allow(dead_code)]
```
irakoton, line 50: I'm a bit confused, weren't labels, before your PR, used precisely in the verifier gadget? How come it's dead code now?

irakoton (approved):
Great PR, thanks for the work. A few comments.

proofs/src/plonk/circuit.rs
miguel-ambrona: This was intentional. We had 2 functions almost identical, I decided to implement one out of the other:
https://github.com/midnightntwrk/midnight-zk/pull/392/changes/c6d3f9fdcfe64cdeb5d58357bdaf4ae5c8b6ddeb

This PR is on a general refactor, not  limited to labeling all polys.

proofs/src/utils/helpers.rs
miguel-ambrona: Yes, that may be better.

At some point I even considered not giving the lookup a name but just a `usize`. That should work too.

proofs/src/poly/commitment.rs
miguel-ambrona: I also dislike it, but it is what we agreed after a discussion with Iñigo.

Apart from serializing the labels (which is handy, but not very elegant and it feels wasteful) we could also revisit the signature of `read` to require the labels, but that is even worse (since `read` is from a global trait used elsewhere).

I think the plan is to stick to this. @iquerejeta ?

proofs/src/poly/query.rs
miguel-ambrona: Makes sense.

proofs/src/poly/query.rs
miguel-ambrona: Yes, I'm a bit skeptical about Custom, but it seems useful.
Currently only used for the multi-open  internal commitments.

circuits/src/verifier/accumulator.rs
miguel-ambrona: Well, no. Only if it is a fixed base.
But this is gonna change a lot. The abstraction in the verifier gadget is wrong. Let's open this conversation after I apply the refactor I have in mind.

circuits/src/verifier/mod.rs
miguel-ambrona: They are not used in the verifier gadget. I only added them for consistency.
This will be addressed in the future.

proofs/src/plonk/mod.rs
```diff
@@ -177,7 +180,10 @@ where
         let num_fixed_columns = u32::from_le_bytes(num_fixed_columns);
 
         let fixed_commitments: Vec<_> = (0..num_fixed_columns)
-            .map(|_| CS::Commitment::read(reader, format))
+            .map(|i| {
+                CS::Commitment::read(reader, format)
+                    .map(|c| c.label(PolynomialLabel::Fixed(i as usize)))
+            })
```
iquerejeta, line 186: Before you were labelling the generator with `i = 0`. Won't there be a collision between the generator and the first fixed poly?

proofs/src/poly/kzg/mod.rs
```diff
@@ -210,7 +210,7 @@ where
             poly_inner_product(&f_polys, powers(x2))
         };
 
-        let f_com = Self::commit(params, &f_poly, PolynomialLabel::NoLabel);
+        let f_com = Self::commit(params, &f_poly, PolynomialLabel::Custom("kzg_batch".into()));
```
iquerejeta: what about the following name? 
```suggestion
        let f_com = Self::commit(params, &f_poly, PolynomialLabel::Custom("multi_open_batch".into()));
```

proofs/src/poly/kzg/commitment.rs
```diff
@@ -78,15 +78,28 @@ where
...
+impl<E: MultiMillerLoop> Labelable for KZGCommitment<E> {
+    fn label(self, label: PolynomialLabel) -> Self {
+        match self {
+            Self::Simple(p, _) => Self::Simple(p, label),
+            Self::Linear(points, scalars, labels) => Self::Linear(
+                points,
+                scalars,
+                labels.into_iter().map(|_| label.clone()).collect(),
+            ),
+        }
```
iquerejeta, line 96: Do we ever 'relable'? If not, I would check that the label matches the existing one in case it is not `None`.

iquerejeta (dismissed):
Looks good to me - minor comments.

irakoton (dismissed):
I was coding on top of the PR, and I encountered an error that isn't caught by the CLI (due to being triggered only when truncated challenge is enabled). Confirmed the error doesn't trigger on main but triggers on your branch, so I think it comes from here.

Reproduce with
```bash
cargo test -p midnight-proofs --lib --features truncated-challenges poly::kzg::tests::test_roundtrip_gwc
```

I'm not exactly sure what's going on, but it seems there is a NoLabel running wild in the MSM.

proofs/src/poly/kzg/commitment.rs
miguel-ambrona: Good point. I will check the label is `None` on Simple commitments and I will not allow labeling Linear ones.

proofs/src/plonk/mod.rs
miguel-ambrona: I'm not sure why you say we were labeling the generator with `i = 0`.

The (opposite of the) generator should have label `Custom("-G")`, what do you think?

miguel-ambrona:
> I was coding on top of the PR, and I encountered an error that isn't caught by the CLI (due to being triggered only when truncated challenge is enabled). Confirmed the error doesn't trigger on main but triggers on your branch, so I think it comes from here.
> 
> Reproduce with
> 
> ```shell
> cargo test -p midnight-proofs --lib --features truncated-challenges poly::kzg::tests::test_roundtrip_gwc
> ```
> 
> I'm not exactly sure what's going on, but it seems there is a NoLabel running wild in the MSM.

Solved, this was due to the addition of new labels.

proofs/src/plonk/mod.rs
iquerejeta: You are right - got confused with this instance commitment [here](https://github.com/midnightntwrk/midnight-zk/pull/392/changes/c9017fc4a9ee8c01c570ab9d0e9dfb15e5b0d01d#diff-41438af180d2504e9d1304c7e3ad356db11393934d7d4d50ffd67b2e67679e3eR355)

iquerejeta (approved):

PR #408, opened 2026-06-04
Release new version v2.3

miguel-ambrona:
Introduce Instantiable::from_public_input as a decoding function and (injectivity tests).

Expose conversion functions between scalars and biguint in native Edwards chip.


iquerejeta (approved):
Looks good to me

PR #411, opened 2026-06-05
Remove `Query` trait

miguel-ambrona:
Removes the `Query` trait from `poly/query.rs` and replaces its role in `construct_intermediate_sets` with a concrete `&[(T, F, F)]` tuple interface (commitment reference, point, eval).

### DIFF
--- a/proofs/src/poly/query.rs
+++ b/proofs/src/poly/query.rs
@@ -2,10 +2,7 @@ use std::fmt::{self, Debug};
 
 use ff::PrimeField;
 
-use crate::{
-    poly::{commitment::PolynomialCommitmentScheme, Coeff, Polynomial},
-    utils::arithmetic::eval_polynomial,
-};
+use crate::poly::{commitment::PolynomialCommitmentScheme, Coeff, Polynomial};
 
 /// A structured label for polynomial commitments in verifier queries.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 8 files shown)


proofs/src/poly/query.rs
```diff
@@ -89,34 +77,20 @@ where
...
+        }
     }
 }
 
 #[doc(hidden)]
 #[derive(Copy, Clone, Debug)]
-pub struct PolynomialPointer<'com, F: PrimeField> {
-    pub(crate) poly: &'com Polynomial<F, Coeff>,
-}
+pub struct PolynomialReference<'com, F: PrimeField>(pub(crate) &'com Polynomial<F, Coeff>);
```
davidnevadoc, line 89: Not an addition of this PR really, but I wonder why do we need this struct.
In particular, why do we hardcode `Coeff`? It is not better for computing evaluations than other basis.

davidnevadoc (approved):
I like any and all efforts to try to simplify this part of the code. But I'm not sure if we are really achieving that.
On the one hand, the unnamed tuple `[(T, F, F)]` makes it even harder to follow what is going on (despite the nice comment). But on the other hand, there are already too many structs and traits that do not have a clearly justified purpose ( `Polynomialreference`, `CommitmentReference`, `CommitmentData`, `IntermediateSets` ... way too much stuff for something that should be simpler.

I think it would be good to think of a bigger refactor of this module. And also, to wait for fflonk to be merged to avoid complicating that workstream even more.

miguel-ambrona:
> I like any and all efforts to try to simplify this part of the code. But I'm not sure if we are really achieving that.

The `Query` trait was a blocker for future refactors I'm planning.

> And also, to wait for fflonk to be merged to avoid complicating that workstream even more.

Fflonk cannot be merged without the refactor I'm doing. It is not that simple.

proofs/src/poly/query.rs
miguel-ambrona: It is there to make sure we compare polynomials by reference instead of by value when building the intermediate sets.

proofs/src/poly/query.rs
```diff
@@ -89,34 +77,20 @@ where
...
-    type Eval = F;
-
-    fn get_point(&self) -> F {
-        self.point
-    }
-    fn get_eval(&self) -> Self::Eval {
-        eval_polynomial(&self.poly[..], self.get_point())
-    }
-    fn get_commitment(&self) -> Self::Commitment {
-        PolynomialPointer { poly: self.poly }
```
davidnevadoc, line 119: 🤯

iquerejeta (approved):

PR #412, opened 2026-06-08
Refactor `curves`

### LINKED ISSUE #366: README of curves is outdated
Just realised, when publishing v0.3.0, that the README of curves crate is outdated.

### Documentation URL:
https://crates.io/crates/midnight-curves

### Description of Improvement:
We should adapt to the new crate structure.

davidnevadoc:
## Summary

   - Remove dead/duplicated code:
      Dead CurveExt methods, the hash-to-curve subsystem (off-circuit  hash-to-curve already lives in midnight-circuits), the redundant serde derive layer, and unused  features/deps.
   - Reorganize the `dev-only` bn256 internals:
   - Simplify the serialization surface
   - Updates curves README (closes #366 )

 ## Breaking changes (public API)

  1. `serde::{Serialize, Deserialize}` impls removed for all field/curve types, and the serde cargo feature removed → use `group::GroupEncoding` (compressed) or `SerdeObject` (raw).
  2. Removed BLS  `helpers unique_messages`, `PairingG1G2`, `PairingG2G1`.
  3. Removed unused `CurveExt` methods `endo`, `jacobian_coordinates`, `new_jacobian`, `hash_to_curve` and `Coordinates::{u, v}`.
  4. Removed the `hash_to_curve` module and BLS `G1/G2::hash_to_curve` inherent methods.
  5. Moved the generic extension-field tower (ExtField, quadratic/cubic) from ff_ext into the dev-curves bn256 module.
  6. Removed the `__private_bench` feature (and its Fp12/Fp2 re-exports) and the unused halo2curves workspace dependency.


curves/src/tests/curve.rs
```diff
@@ -88,12 +88,12 @@ macro_rules! curve_testing_suite {
...
                 let a = $c::identity();
                 let a = a.double();
-                assert!(bool::from(c.is_on_curve()));
+                assert!(bool::from(a.is_on_curve()));
                 assert!(bool::from(a.is_identity()));
 
                 let a = $c::generator();
                 let a = a.double();
-                assert!(bool::from(c.is_on_curve()));
+                assert!(bool::from(a.is_on_curve()));
```
miguel-ambrona, line 96: Was this wrong before?

miguel-ambrona (dismissed):
I mainly reviewed the insertions and code preservation, not so much the deletions.

Shall we mention in the relevant CHANGELOG that the `derive_serde` feature has been removed?

Is `digest` in `curves/Cargo.toml` now redundant?

curves/src/bn256/fq2.rs
```diff
@@ -5,19 +5,195 @@ use ff::{Field, FromUniformBytes, PrimeField, WithSmallOrderMulGroup};
...
+    fn from_repr(repr: Self::Repr) -> CtOption<Self> {
+        let c0: [u8; Fq::SIZE] = repr[..Fq::SIZE].try_into().unwrap();
+        let c0: <Fq as PrimeField>::Repr = c0.into();
+        let c0 = Fq::from_repr(c0);
+
+        let c1: [u8; Fq::SIZE] = repr[Fq::SIZE..].try_into().unwrap();
+        let c1: <Fq as PrimeField>::Repr = c1.into();
+        let c1 = Fq::from_repr(c1);
+
+        CtOption::new(Fq2::new(c0.unwrap(), c1.unwrap()), Choice::from(1))
```
remalue: Not related to your refactor, but can a panic from this unwrap cause any problem?

remalue:
LGTM, as far as I can tell.

curves/CHANGELOG.md
```diff
@@ -13,9 +13,10 @@ and this project adheres to [Semantic Versioning](https://book.async.rs/overview
 * Add `k256` module [#189](https://github.com/midnightntwrk/midnight-zk/pull/189), [#191](https://github.com/midnightntwrk/midnight-zk/pull/191)
 
 ### Changed
-
+* BREAKING: Moved the generic extension-field tower (`ExtField`, `quadratic`/`cubic`) from `ff_ext` to the dev-curves `bn256` module [#412](https://github.com/midnightntwrk/midnight-zk/pull/412)
```
irakoton: What is the semantics of `BREAKING`? Is this an official keyword we have to use when writing changelogs, or is this just you here happening to mention it occasionally?

Cargo.toml
```diff
@@ -8,7 +8,6 @@ package = { license-file = "LICENSE" }
 [workspace.dependencies]
 ff = "0.13"
 group = "0.13"
-halo2curves = { version = "0.7.0", default-features = false }
```
irakoton, line 11: I thought we had removed dependencies to halo2curves a while ago, I even think I took a look at the PR back then. Was this something else, or related and this PR is the finishing touch? Or am I just crazy

curves/src/bn256/fq2.rs
```diff
@@ -5,19 +5,195 @@ use ff::{Field, FromUniformBytes, PrimeField, WithSmallOrderMulGroup};
...
+        let c0 = Fq::from_bytes(bytes[0..Fq::SIZE].try_into().unwrap());
+        let c1 = Fq::from_bytes(bytes[Fq::SIZE..Fq::SIZE * 2].try_into().unwrap());
+        CtOption::new(
+            Fq2 {
+                c0: c0.unwrap(),
+                c1: c1.unwrap(),
+            },
+            c0.is_some() & c1.is_some(),
+        )
+    }
```
irakoton: It's arguably minor, but the way the options are managed is weird. First they are unwrapped, and then `is_some` is called.  You could just write `true` for the boolean.
Also, what is the reason for using `CtOption`, is there any need for constant time in the library?

irakoton: I'm not sure if the fix you want is
1. hardcoding `true`
2. using `and_then` + `zip` on `c0` and `c1`

curves/src/bn256/fq2.rs
```diff
@@ -5,19 +5,195 @@ use ff::{Field, FromUniformBytes, PrimeField, WithSmallOrderMulGroup};
...
+
+        CtOption::new(Fq2::new(c0.unwrap(), c1.unwrap()), Choice::from(1))
+    }
+
+    fn to_repr(&self) -> Self::Repr {
+        let mut res = Self::Repr::default();
+        let c0 = self.c0.to_repr();
+        let c1 = self.c1.to_repr();
+        res[0..Fq::SIZE].copy_from_slice(&c0.as_ref()[..]);
+        res[Fq::SIZE..Fq::SIZE * 2].copy_from_slice(&c1.as_ref()[..]);
```
irakoton: I get a clippy warning for these full-range slices.

curves/src/bn256/engine.rs
```diff
@@ -11,13 +11,255 @@ use pairing::{Engine, MillerLoopResult, MultiMillerLoop, PairingCurveAffine};
...
+    }
+
+    /// Doubles this group element.
+    pub fn double(&self) -> Gt {
+        use ff::Field;
+        Gt(self.0.square())
+    }
+}
+
+impl<'a> Neg for &'a Gt {
```
irakoton: clippy warning for elidable `'a`.

curves/src/bn256/engine.rs
```diff
@@ -11,13 +11,255 @@ use pairing::{Engine, MillerLoopResult, MultiMillerLoop, PairingCurveAffine};
...
+impl Neg for Gt {
+    type Output = Gt;
+
+    #[inline]
+    fn neg(self) -> Gt {
+        -&self
+    }
+}
+
+impl<'a, 'b> Add<&'b Gt> for &'a Gt {
```
irakoton: clippy warning for elidable `'a` (not `b`).

curves/src/bn256/engine.rs
```diff
@@ -11,13 +11,255 @@ use pairing::{Engine, MillerLoopResult, MultiMillerLoop, PairingCurveAffine};
...
+    type Output = Gt;
+
+    #[inline]
+    #[allow(clippy::suspicious_arithmetic_impl)]
+    fn add(self, rhs: &'b Gt) -> Gt {
+        Gt(self.0 * rhs.0)
+    }
+}
+
+impl<'a, 'b> Sub<&'b Gt> for &'a Gt {
```
irakoton: clippy warning for elidable `'a` (not `b`).

curves/src/bn256/engine.rs
```diff
@@ -11,13 +11,255 @@ use pairing::{Engine, MillerLoopResult, MultiMillerLoop, PairingCurveAffine};
...
+    type Output = Gt;
+
+    #[inline]
+    fn sub(self, rhs: &'b Gt) -> Gt {
+        self + (-rhs)
+    }
+}
+
+#[allow(clippy::suspicious_arithmetic_impl)]
+impl<'a, 'b> Mul<&'b Fr> for &'a Gt {
```
irakoton: clippy warning for elidable `'a` (not `b`).

curves/src/bn256/fq2.rs
irakoton: I think there is at least the issue that the convention is not consistent with the one used at L58. I would either do always like L137 (panic + hardcode `true`), or cleanly return `none` if the representation is not canonical (i.e., use `and_then` + `zip` on `c0` + `c1`, and use `is_some` for the boolean flag). Since we're returning an option anyway, the second solution is probably cleaner.

curves/src/bn256/fq2.rs
```diff
@@ -5,19 +5,195 @@ use ff::{Field, FromUniformBytes, PrimeField, WithSmallOrderMulGroup};
...
+        self.c0.write_raw(writer)?;
+        self.c1.write_raw(writer)
+    }
+}
+
+impl FromUniformBytes<96> for Fq2 {
+    fn from_uniform_bytes(bytes: &[u8; 96]) -> Self {
+        const SIZE: usize = 96 / 2;
+        let c0: [u8; SIZE] = bytes[SIZE..].try_into().unwrap();
+        let c1: [u8; SIZE] = bytes[..SIZE].try_into().unwrap();
```
irakoton, line 182: Is this a typo? Why is `c0` the second part of `bytes` and `c1` the first part?

irakoton (changes requested):
Nice refactor, but there are a couple of nits + weird managements of Options that need to be handled I think. 

~I would like to make another pass, in particular in the tests, because the diffs are quite big and I'm not confident the coverage is still preserved.~
Added comments.

curves/src/bn256/engine.rs
```diff
@@ -207,14 +449,10 @@ fn ell(f: &mut Fq12, coeffs: &(Fq2, Fq2, Fq2), p: &G1Affine) {
...
+    use super::super::Bn256;
 
-    use super::{
-        super::{Bn256, Fr, G1, G2},
-        multi_miller_loop, Fq12, G1Affine, G2Affine, Gt,
-    };
-    crate::test_pairing!(Bn256, G1, G1Affine, G2, G2Affine, Fq12, Gt, Fr);
+    #[test]
+    fn bn256_engine_tests() {
+        crate::tests::engine::engine_tests::<Bn256>();
```
irakoton, line 456: Is there a reason why only `engine_tests` is called, and not all the test functions that were present in `test_pairing!`? Like `test_bilinearity` and a couple of others.

curves/src/bn256/curve_impl.rs
```diff
@@ -57,31 +52,31 @@ macro_rules! new_curve_impl {
...
 
                 fn from_bytes(bytes: &Self::Repr) -> CtOption<Self> {
-                    <Self::Repr as $crate::serde::Compressed<$name_affine>>::decode(bytes.clone()).map(Self::from)
+                    <Self::Repr as $crate::bn256::compressed::Compressed<$name_affine>>::decode(bytes.clone()).map(Self::from)
                 }
 
                 fn from_bytes_unchecked(bytes: &Self::Repr) -> CtOption<Self> {
-                    <Self::Repr as $crate::serde::Compressed<$name_affine>>::decode(bytes.clone()).map(Self::from)
+                    <Self::Repr as $crate::bn256::compressed::Compressed<$name_affine>>::decode(bytes.clone()).map(Self::from)
                 }
```
irakoton, line 76: Well it's not really specific to this PR, and there are other places where this is the case, but I'm a bit confused that `from_bytes` and `from_bytes_unchecked` have exactly the same code. Is this this whole debate about subgroup checks being useless? Why not inlining a call to `from_bytes` in the `_unchecked` version for clarity then?

curves/src/tests/curve.rs
davidnevadoc: yes, this was wrong

curves/src/bn256/curve_impl.rs
davidnevadoc: `GroupEncoding` provides these to versions of deserialization (checked and unchecked).
However, the compression (or rather de-compression) makes the curve-check implicit, so the only check to skip would be the subgroup memebership, which is only needed in non-prime order curves.
As a result, for bn256 both functions are the same. They are not for bls12-381 for example.

> Why not inlining a call to from_bytes in the _unchecked version for clarity then?

Hmmm, since both are 1-line I find it clearer to have the same body, rather than a call.

curves/src/bn256/fq2.rs
davidnevadoc: Good catch.
I have used `to_bytes` and `from_bytes` in `to_repr` and `from_repr` for consistency.

Cargo.toml
davidnevadoc: It was removed from `curves`. I think this is just a leftover.

curves/src/bn256/fq2.rs
davidnevadoc: Yes, this is quite messy. It has been fixed with a similar comment in `from_repr`

curves/src/bn256/fq2.rs
davidnevadoc: :man_facepalming: This is a mess... 
I think this trait shouldn't even be implemented. It is a super trait of `PrimeFiel` which of course shouldn't be implemented for Fq2 either.
The appropriate thing should be to eliminate it altogether. Unless we do that I'd keep it as is to maintain behaviour.

curves/src/bn256/fq2.rs
davidnevadoc: All these clippy warnings where a result of unrolling the macos without cleaning up the resulting code.
They have all been fixed

curves/src/bn256/engine.rs
davidnevadoc: Fixed

curves/src/bn256/engine.rs
davidnevadoc: Fixed

curves/src/bn256/engine.rs
davidnevadoc: Fixed

curves/src/bn256/engine.rs
davidnevadoc: Now we call the tests that are implemented generically for `Engine`, which cover everything that was in `test_pariring!`. In particular, that is covered by `fn random_bilinearity_tests`(which covers more additional cases).

curves/src/bn256/engine.rs
davidnevadoc: fixed

curves/CHANGELOG.md
davidnevadoc: Not really, I just used it because I was trying to keep track of the API breaking changes. I will revert to standard formatting

irakoton (dismissed):
All good now!

iquerejeta (dismissed):

iquerejeta (dismissed):

iquerejeta (approved):

PR #417, opened 2026-06-10
Bump `sha3-circuit`, `blake2b_halo2` version

davidnevadoc:
External crates dependencies `blake2b_halo2` and `sha3-circuit` have been updated from the previous pinned github rev, to the current crates.io latest version.

WIP: There is an issue with `proofs` and `curves` patches and breaking compilation


iquerejeta:
Why are you doing this? 0.2.0 is behind that commit 🤔

PR #425, opened 2026-06-17
Fix G1 `from_uncompressed` missing check

### LINKED ISSUE #426: G1 `from_uncompressed` skips the prime-order subgroup check
## Summary
  `G1Affine::from_uncompressed` validates only
  `is_on_curve()` and omits `is_torsion_free()`. It therefore accepts on-curve G1
  points that lie **outside the prime-order subgroup**. This is the single chokepoint
  for the checked uncompressed / `SerdeFormat::RawBytes` G1 reads — `from_raw_bytes`,
  `read_raw`, and `UncompressedEncoding::from_uncompressed` all route through it.

  G1 *compressed* (`from_compressed`) and **both** G2 paths already perform the check,
  so this is an inconsistency in one path, not a systemic gap.

  ## Impact
  - **External API :** a "checked" deserializer silently performs no subgroup validation. For external consumers of the published `midnight-curves` crate, ingesting untrusted G1 bytes via `RawBytes` is a latent vulnerability.
  - A `VerifyingKey`/`ProvingKey`/`ParamsKZG` read with `RawBytes` accepts a non-subgroup G1 commitment/base.

  ## What is NOT affected
  - **Proof verification soundness:** a stray cofactor component is *pairing-invisible*
    (G1 cofactor order is coprime to `r`), so it contributes a factor of 1 to every KZG/pairing check, it cannot satisfy a false equation.
  - **Proof commitments / transcript:** read **compressed + checked**  (`GroupEncoding::from_bytes`), which always validated. Unaffected.
  - **SRS download path:** uses `RawBytesUnchecked` (no check by design), so this fix neither changes nor protects it.
  - **Availability:** a non-subgroup vk point does **not** panic/crash the verifier.
    `blst` arithmetic + pairing are domain-general ; the proof just fails verification gracefully.
  - **Stored data / compatibility:** nothing ever *emits* a non-subgroup point, so all
    honestly-serialized artifacts still round-trip. Not a breaking change for valid data.

  ## Fix
  Add `& p.is_torsion_free()` to `from_uncompressed`.
  Fixed in #425

davidnevadoc:
## Changes

  - Add the prime-order subgroup check (`is_torsion_free()`) to `G1Affine::from_uncompressed`.
  - Drop the now-redundant `is_on_curve()` from the 4 checked G1/G2 deserializers (blst already guarantees
  on-curve there); each collapses to just `is_torsion_free()`.
  - Add G1/G2 tests asserting every checked path (`from_uncompressed`, `from_compressed`, `from_raw_bytes`,
  `read_raw`, `UncompressedEncoding`) rejects a non-subgroup point while `_unchecked` accepts it.
  
  
  Fixes #426


iquerejeta (commented):
Are we sure that the check 'is_torsion_free' checks that it is indeed in the curve? 

see [here](https://github.com/supranational/blst/blob/d29fef46c29a662ec215a9eeb9578e25f9146b6d/src/map_to_g1.c#L531)

curves/src/bls12_381/g1.rs
```diff
@@ -1576,6 +1576,61 @@ mod tests {
...
+    /// prime-order subgroup. The explicit `is_torsion_free` guard makes the
+    /// helper correct even in the astronomically unlikely case it lands in it.
+    fn g1_non_torsion_point() -> G1Affine {
+        let b = Fp::from(4u64);
+        let mut x = Fp::ONE;
+        loop {
+            let rhs = x.square() * x + b;
+            if let Some(y) = Option::<Fp>::from(rhs.sqrt()) {
+                let p = G1Affine::from_raw_unchecked(x, y, false);
+                if bool::from(p.is_on_curve()) && !bool::from(p.is_torsion_free()) {
```
iquerejeta: `is_on_curve` is already performed in `from_raw_unchecked`, right?

iquerejeta: This also applies to G2. And for the same reason, I don't think we need to do the checks below that `assert_eq!(p.is_on_curve().unwrap_u8(), 1);`

iquerejeta (changes requested):
Can we add these checks to the function `from_xy` as well?

curves/src/bls12_381/g1.rs
davidnevadoc: They are not necessary, but I think it is good to have explicit calls to understand what the test is checking very specifically.

davidnevadoc:
> Can we add these checks to the function `from_xy` as well?

There are two `from_xy`: in `CircuitCurve` and in `CurveAffine`. 
`CircuitCurve` is fine as is. It does not imply prime group memebership. That is why it has the associated type `CryptpgraphicGroup` and the function `into_subgroup()` that removes the cofactor if necessary and returns an element guaranteed to be in the subgroup.

`CurveAffine` is not that clear. `G1Affine` implements `PrimeCurveAffine`  and so does `G2Affine`.... so I'm not sure what to make of this. Its probably best to just get rid of these traits

iquerejeta (approved):

PR #427, opened 2026-06-17
Remove the light aggregator

miguel-ambrona:
We remove this module because we have no plans of using it and it is hard to maintain, especially after an ongoing refactor of circuits for the integration of _Fflonk_.
  
If we ever need it, we can always add it back. In that case, it is important to note that the combination of the following two properties is a soundness issue for the light aggregator. Fixing either one would be sufficient:

- `proofs/` verifiers should check that the public input length is the expected one; this check is currently enforced at a higher layer, in `zk_stdlib::verify`, because it is important for Midnight, but not in `proofs/`,

- the number of bases in the accumulator should not be read from the transcript by the verifier, but hardcoded instead.


iquerejeta (approved):
Lovely

PR #430, opened 2026-06-18
Sync verifier gadget with the verifier in `proofs/`

miguel-ambrona:
- Introduce `AssignedKZGCommitment` as the in-circuit analog of `KZGCommitment`, replacing the previous `LabeledPoint` type.

- Replace the string-based fixed base naming scheme with `PolynomialLabel`s.

- Get rid of the `fixed_base_scalars` map in MSMs by introducing `Point` and `AssignedPoint` enums in `verifier::msm` to distinguish globally-fixed bases from variable ones.

- Restructure `Msm` and `AssignedMsm` as a flat triple (`scalars`, `bases`, `labels`). Just like in `proofs/`.

- Finally add the IVC example to the CI.

> [!NOTE]  
> I recommend reviewing most commits by comparing the new verifier gadget code with the one in `proofs/verifier`. Rather than trying to understand the diff.


circuits/src/verifier/msm.rs
```diff
@@ -37,146 +38,280 @@ use crate::{
...
+        match point {
+            Point::Variable(p) => S::AssignedPoint::as_public_input(p),
+            Point::Fixed => vec![],
+        }
+    }
+
+    #[cfg(any(test, feature = "testing"))]
+    fn from_public_input(fields: &[S::F]) -> Option<Point<S>> {
+        match fields {
+            [] => Some(Point::Fixed),
```
iquerejeta, line 109: This looks odd (also in the function `as_public_input`). Seems like one would only call this function by mistake - why not panic? Or am I missing an actual use case for this?

circuits/src/verifier/msm.rs
```diff
@@ -37,146 +38,280 @@ use crate::{
...
+        terms: &[(&PolynomialLabel, &S::F, &S::C)],
+        fixed_bases: &BTreeMap<PolynomialLabel, S::C>,
+    ) -> Self {
+        let mut bases = Vec::with_capacity(terms.len());
+        let mut scalars = Vec::with_capacity(terms.len());
+        let mut labels = Vec::with_capacity(terms.len());
+
+        for (label, scalar, &base) in terms {
+            scalars.push(**scalar);
+            labels.push((*label).clone());
```
iquerejeta, line 199: What about implementing `Copy` for polynomial label to avoid these clones? I think it can easily be derived.

circuits/src/verifier/msm.rs
```diff
@@ -390,102 +546,115 @@ impl<S: SelfEmulation> AssignedMsm<S> {
...
+        self.scalars = scalars;
+        self.labels = labels;
 
         Ok(())
     }
 
-    /// Given the actual fixed bases, resolves the fixed-base part of the MSM
-    /// by pairing each named scalar with its base and moving them to regular
-    /// variable-base entries.
+    /// Resolves all `Fixed` bases by substituting their assigned circuit
```
iquerejeta, line 635: Mention in this comment that this is the in-circuit analog of `resolve_fixed_bases`

proofs/src/poly/query.rs
```diff
@@ -5,7 +5,7 @@ use ff::PrimeField;
 use crate::poly::{commitment::PolynomialCommitmentScheme, Coeff, Polynomial};
 
 /// A structured label for polynomial commitments in verifier queries.
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
iquerejeta, line 8: How is `PartialOrd` derived here? e.g., how are `::Fixed` and `::Advice` compared?

circuits/src/verifier/transcript_gadget.rs
```diff
@@ -97,13 +98,23 @@ impl<S: SelfEmulation> TranscriptGadget<S> {
...
     }
 
-    /// Absorbs a point into the transcript.
-    pub fn common_point(
+    /// Absorbs a commitment into the transcript.
+    pub fn common_commitment(
         &mut self,
         layouter: &mut impl Layouter<S::F>,
-        point: &S::AssignedPoint,
+        point: &AssignedKZGCommitment<S>,
```
iquerejeta, line 105: rename to 'commitment'?

circuits/src/verifier/msm.rs
```diff
@@ -314,73 +475,68 @@ impl<S: SelfEmulation> AssignedMsm<S> {
...
-    /// The points of the MSM are not enforced to be part of the relevant prime
-    /// order subgroup.
+    /// Variable-base points are not checked for subgroup membership.
     ///
     /// # Panics
     ///
-    /// If `msm_value` is known and its number of variable bases differs from
-    /// `len`, or its number of fixed-base scalars differs from
-    /// `fixed_base_names.len()`.
+    /// If `msm_value` is known and its label list differs from `labels`.
```
iquerejeta, line 502: If I understand this correctly, this is an error, and not a panic. I think that's good, but I would change the docs here, and also in the assignment of the accumulator.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -172,34 +173,32 @@ impl<S: SelfEmulation> VerifierGadget<S> {
...
-    /// Witnesses an accumulator with just 1 non-fixed base-scalar pair on each
-    /// side (the scalar being constant 1), and as many fixed-base scalars (on
-    /// its right-hand-side) as provided through the `fixed_base_names`
-    /// argument (no fixed-base scalars on the left-hand-side).
+    /// Witnesses the "collapsed" form of a KZG accumulator.
+    ///
+    /// The expected shape is:
+    /// * LHS: exactly one `Variable` entry labeled `Collapsed` with scalar `1`.
+    /// * RHS: one `Fixed` entry per label in `fixed_base_labels` plus one
+    ///   `Variable` entry labeled `Collapsed` with scalar `1`.
```
iquerejeta, line 181: I'd be explicit in stating that if this condition is not satisfied, the function errors.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -843,7 +837,7 @@ impl<S: SelfEmulation> VerifierGadget<S> {
             &self.curve_chip,
             &self.scalar_chip,
             &mut transcript,
-            &queries.collect::<Vec<_>>(),
+            &queries,
         )?;
 
         Ok(multiopen_check)
```
iquerejeta: We can probably remove the `multiopen_check` intermediate variable (and mimic what we do in the off-circuit verifier)

circuits/src/verifier/kzg.rs
iquerejeta: We could use a fold here, like in the off-circuit verifier

circuits/src/verifier/kzg.rs
```diff
@@ -460,9 +607,9 @@ pub(crate) fn multi_prepare<S: SelfEmulation>(
 
     let right = {
         let mut right = final_com; // C
-        let minus_v_gen = AssignedMsm::from_fixed_term(&v, "-G");
-        right.add_msm(layouter, scalar_chip, &minus_v_gen)?; // -vG
-        right.add_msm(layouter, scalar_chip, &scaled_pi)?; // zπ
+        let minus_v_gen = AssignedMsm::from_fixed_term(v, PolynomialLabel::Custom("-G".into()));
+        right.add_msm(&minus_v_gen)?; // -vG
+        right.add_msm(&scaled_pi)?; // zπ
```
iquerejeta, line 620: Seems slightly off-sync with the off-circuit verifier. Maybe it makes more sense to change things in the other verifier, where instead of creating the `DualMSM` so early, we just create it when return it.

iquerejeta (dismissed):
LGTM - minor comments.

circuits/src/verifier/kzg.rs
miguel-ambrona: I'll open an issue, I do not wanna change this now.

circuits/src/verifier/transcript_gadget.rs
miguel-ambrona: No, this will be generalized in a future PR.

circuits/src/verifier/kzg.rs
davidnevadoc: How is this off-sync?

iquerejeta: Just the code, not the logic.

circuits/src/verifier/msm.rs
miguel-ambrona: This is the exact inverse of `as_public_input`. This is currently correct.

But you are right to question this. Maybe we should add a public input field element e.g. zero every time `as_public_input` is called on a `Point::Fixed`.

circuits/src/verifier/msm.rs
miguel-ambrona: Copy cannot be derived because the label uses strings.

proofs/src/poly/query.rs
miguel-ambrona: Does it matter? All we need is an order.

davidnevadoc: I think it doesn't matter which specific order we choose but it seems a good idea to have the order explicitly specified. Even if it is just for ease of debugging.

miguel-ambrona: The derived order is pretty clear, it follows the order of enum items and calls `PartialEq` of the internal data to break a tie if necessary.

iquerejeta: Wasn't clear to me - but I see that it is indeed well defined for enums. I'm good to keep it as is.

iquerejeta (approved):

scottbuckel (approved):

PR #440, opened 2026-06-23
Generalize `cost_model` to depend on a `PCS`

miguel-ambrona:
`CircuitModel` and its associated cost functions now take the polynomial commitment scheme as a type parameter instead of a hardcoded commitment byte-length.

A commitment_byte_length method is added to `PolynomialCommitmentScheme`.

### DIFF
--- a/proofs/src/dev/cost_model.rs
+++ b/proofs/src/dev/cost_model.rs
@@ -285,8 +285,7 @@ pub fn circuit_model_with<F: Ord + Field + FromUniformBytes<64>>(
 
 /// Given a Plonk circuit, this function returns a [CircuitModel].
 ///
-/// Commitment byte sizes and scalar field element sizes are both derived from
-/// `CS`; no size constants need to be supplied.
+/// Commitment and scalar field sizes (in bytes) are both derived from `CS`.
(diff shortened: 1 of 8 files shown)


proofs/src/dev/cost_model.rs
```diff
@@ -193,120 +194,104 @@ pub struct CircuitModel {
...
-/// Given a Plonk circuit, this function returns a [CircuitModel]
-pub fn circuit_model<
-    F: Ord + Field + FromUniformBytes<64>,
-    const COMM: usize,
-    const SCALAR: usize,
->(
+/// Given a Plonk circuit, this function returns a [CircuitModel].
+///
+/// Commitment byte sizes and scalar field element sizes are both derived from
+/// `CS`; no size constants need to be supplied.
```
iquerejeta:
```suggestion
/// Commitment byte sizes and scalar field element sizes are both derived from
/// `CS`.
```

iquerejeta (dismissed):
Looks good - it just came to mind that it'll be a mess to calculate the number of evaluations with FFlonk. We'll get there.

iquerejeta (approved):

PR #441, opened 2026-06-23
Generalize the verifier gadget to any PCS

miguel-ambrona:
Add an in-circuit Polynomial Commitment Scheme abstraction mirroring `midnight_proofs::poly::commitment` for the in-circuit verifier:

  - `InCircuitHomomorphicCommitment<S>`: in-circuit `mul`/`add` on an additively homomorphic commitment.
  - `InCircuitPCS<S>`: the scheme itself (associated commitment type, `fixed_commitment`, `assign_commitment`, `read_commitment`, `common_commitment`).

### DIFF
--- a/circuits/src/verifier/pcs.rs
+++ b/circuits/src/verifier/pcs.rs
@@ -1,0 +1,149 @@
+// This file is part of MIDNIGHT-ZK.
+// Copyright (C) Midnight Foundation
+// SPDX-License-Identifier: Apache-2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! In-circuit abstraction layer for Polynomial Commitment Schemes.
+//!
+//! Mirrors [`midnight_proofs::poly::commitment`] for the in-circuit setting.
+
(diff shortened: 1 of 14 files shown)


circuits/src/verifier/pcs.rs
```diff
@@ -0,0 +1,149 @@
...
+
+// ---------------------------------------------------------------------------
+// CommitmentReference  (mirrors proofs/src/poly/query.rs)
+// ---------------------------------------------------------------------------
+
+/// A pointer to an in-circuit commitment, with pointer-based equality.
+///
+/// Two references are equal iff they point to the same allocation, so
+/// commitments are grouped by identity rather than by value.
+pub(crate) struct CommitmentReference<'a, C>(pub(crate) &'a C);
```
iquerejeta, line 39: Realising this now. For the in-circuit case, shouldn't we compare by cell in the circuit?

circuits/src/verifier/pcs.rs
```diff
@@ -0,0 +1,149 @@
...
+        layouter: &mut impl Layouter<S::F>,
+        scalar_chip: &S::ScalarChip,
+        other: Self,
+    ) -> Result<Self, Error>;
+}
+
+/// In-circuit abstraction over a Polynomial Commitment Scheme.
+///
+/// Mirrors [`midnight_proofs::poly::commitment::PolynomialCommitmentScheme`]
+/// for the in-circuit verifier.
```
iquerejeta: The 'mirrors' here is a bit confusing, because the interface doesn't look much like the off-circuit one.

circuits/src/verifier/pcs.rs
miguel-ambrona: In-circuit it does not matter as much.

Nevertheless, this changes in the next PR, where I removed the commitment reference from everywhere (and start comparing things by polynomial labels).

iquerejeta (dismissed):

davidnevadoc (approved):
Re-approve as the previous review went stale

PR #450, opened 2026-06-26
Multi-Polynomial Commitment Schemes

miguel-ambrona:
Adds multi-polynomial KZG commitments (`commit_many` / `KZGMultiCommitment`) and moves the prover/verifier to identifying and grouping commitment queries by `PolynomialLabel` rather than by ad-hoc references. Removes the wrapper types and redundant label variant that this makes unnecessary.

## Changes

 - **Multi-polynomial KZG commitments (`commit_many`).** A single commitment can now cover several polynomials, we use labels on the polynomials to be able to refer to them in the queries.

- **Identify multi-open queries by label, group by label.** Queries are keyed by `PolynomialLabel` and grouped by it in the multi-open path, replacing reference-based identification.

- **Drop `PolynomialReference` and `CommitmentReference` wrappers.** No longer needed once queries are label-identified.

- **Add a label parameter to `collapse`.** The collapse operation now carries an explicit `PolynomialLabel`, which will be the label of the collapsed commitment.

- **Make Trash and Logup labels positional.** These labels are now positional rather than named.

- **Drop the redundant `PolynomialLabel::Instance` variant**, using `CommittedInstance` everywhere.

- **`zk_stdlib::verify` takes the committed instance as a `KZGMultiCommitment`** instead of a `G1Affine`.

### DIFF
--- a/aggregation/examples/single_circuit_aggregation.rs
+++ b/aggregation/examples/single_circuit_aggregation.rs
@@ -21,22 +21,21 @@ mod sha_preimage;
 use std::{collections::BTreeMap, time::Instant};
 
 use ff::Field;
-use group::Group;
 use midnight_aggregation::ivc::{self, IvcCircuit, IvcContext, IvcIO, IvcState, IvcTransition};
 use midnight_circuits::{
     hash::poseidon::{PoseidonChip, PoseidonState},
     instructions::{hash::HashCPU, *},
     types::{AssignedNative, Instantiable},
     verifier::{
-        self, Accumulator, AssignedAccumulator, AssignedKZGCommitment, AssignedVk, BlstrsEmulation,
-        InCircuitKZG, SelfEmulation,
+        self, Accumulator, AssignedAccumulator, AssignedKZGMultiCommitment, AssignedVk,
+        BlstrsEmulation, InCircuitKZG, SelfEmulation,
     },
 };
 use midnight_proofs::{
     circuit::{Layouter, Value},
     plonk::{self, ConstraintSystem, Error},
     poly::{
-        kzg::{commitment::KZGCommitment, params::ParamsVerifierKZG, KZGCommitmentScheme},
+        kzg::{commitment::KZGMultiCommitment, params::ParamsVerifierKZG, KZGCommitmentScheme},
         EvaluationDomain, PolynomialLabel,
     },
     transcript::{CircuitTranscript, Transcript},
@@ -240,8 +239,7 @@ impl IvcTransition for ProofAggregation {
             let dual_msm =
                 plonk::prepare::<F, KZGCommitmentScheme<E>, CircuitTranscript<PoseidonState<F>>>(
                     ctx.vk.vk(),
-                    &[KZGCommitment::Simple(
-                        C::identity(),
+                    &[KZGMultiCommitment::commitment_to_zero(
                         PolynomialLabel::Instance(0),
                     )],
                     &[&statement_pis],
@@ -305,10 +303,11 @@ impl IvcTransition for ProofAggregation {
         )?;
 
         // Verify the inner proof in-circuit.
-        let instance_com = AssignedKZGCommitment::<S>::simple(
-            self.std_lib.bls12_381().assign_fixed(layouter, C::identity())?,
+        let instance_com = AssignedKZGMultiCommitment::commitment_to_zero(
+            layouter,
+            self.std_lib.bls12_381(),
             PolynomialLabel::CommittedInstance(0),
-        );
+        )?;
 
         let inner_proof_acc = self.std_lib.verifier().prepare(
             layouter,
--- a/aggregation/examples/single_circuit_aggregation.rs
+++ b/aggregation/examples/single_circuit_aggregation.rs
@@ -240,7 +240,7 @@ impl IvcTransition for ProofAggregation {
                 plonk::prepare::<F, KZGCommitmentScheme<E>, CircuitTranscript<PoseidonState<F>>>(
                     ctx.vk.vk(),
                     &[KZGMultiCommitment::commitment_to_zero(
-                        PolynomialLabel::Instance(0),
+                        PolynomialLabel::CommittedInstance(0),
                     )],
                     &[&statement_pis],
                     &mut transcript,
--- a/circuits/src/verifier/kzg.rs
+++ b/circuits/src/verifier/kzg.rs
@@ -494,9 +494,9 @@ pub(crate) fn multi_prepare_kzg<S: SelfEmulation>(
         // powers are truncated. Exceptionally, the first one is not collapsed,
         // as the first x4 power is 1.
         #[cfg(feature = "truncated-challenges")]
-        coms.iter_mut()
-            .skip(1)
-            .try_for_each(|com| com.collapse(layouter, curve_chip, scalar_chip))?;
+        coms.iter_mut().skip(1).try_for_each(|com| {
+            com.collapse(layouter, curve_chip, scalar_chip, PolynomialLabel::NoLabel)
+        })?;
         coms.push(f_com.into_msm(layouter, scalar_chip)?);
 
         msm_inner_product(layouter, scalar_chip, &coms, &truncated_x4_powers)?
--- a/circuits/src/verifier/kzg.rs
+++ b/circuits/src/verifier/kzg.rs
@@ -503,7 +503,7 @@ pub(crate) fn multi_prepare_kzg<S: SelfEmulation>(
         for (idx, dummy_point) in dummy_openings {
             queries.push(VerifierQuery {
                 point: dummy_point,
-                commitment_ref: queries[idx].commitment_ref,
+                commitment: queries[idx].commitment,
                 label: queries[idx].label.clone(),
                 eval: transcript_gadget.read_scalar(layouter)?,
             });
@@ -522,7 +522,7 @@ pub(crate) fn multi_prepare_kzg<S: SelfEmulation>(
     let label_to_commitment: HashMap<PolynomialLabel, &AssignedKZGCommitment<S>> = queries
         .iter()
         .map(|q| {
-            let inners = &q.commitment_ref.0 .0;
+            let inners = &q.commitment.0;
             let inner = if inners.len() == 1 {
                 &inners[0]
             } else {
--- a/circuits/src/verifier/kzg.rs
+++ b/circuits/src/verifier/kzg.rs
@@ -27,6 +27,7 @@ use std::{
 };
 
 use ff::Field;
+use group::Group;
 use midnight_proofs::{
     circuit::{Layouter, Value},
     plonk::Error,
@@ -41,7 +42,7 @@ use crate::{
     types::InnerValue,
     verifier::{
         msm::{AssignedMsm, AssignedPoint},
-        pcs::{CommitmentReference, InCircuitHomomorphicCommitment, InCircuitPCS, VerifierQuery},
+        pcs::{InCircuitHomomorphicCommitment, InCircuitPCS, VerifierQuery},
         transcript_gadget::TranscriptGadget,
         utils::{
             evaluate_interpolated_polynomial, inner_product, mul_add, mul_bounded_scalars,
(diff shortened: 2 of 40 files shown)


.github/workflows/aggregation_examples.yml
davidnevadoc, line 1: Does this belong in this PR?

aggregation/examples/single_circuit_aggregation.rs
```diff
@@ -240,9 +239,8 @@ impl IvcTransition for ProofAggregation {
             let dual_msm =
                 plonk::prepare::<F, KZGCommitmentScheme<E>, CircuitTranscript<PoseidonState<F>>>(
                     ctx.vk.vk(),
-                    &[KZGCommitment::Simple(
-                        C::identity(),
-                        PolynomialLabel::Instance(0),
+                    &[KZGMultiCommitment::commitment_to_zero(
```
davidnevadoc, line 242: nice

proofs/src/poly/kzg/commitment.rs
```diff
@@ -34,21 +34,26 @@ pub enum KZGCommitment<E: MultiMillerLoop> {
...
-    /// labels, accumulated during verification for MSM batching.  Never
-    /// serialized or hashed directly.
+    /// labels, accumulated during verification for MSM batching.
+    /// Never serialized or hashed directly.
     Linear(Vec<E::G1>, Vec<E::Fr>, Vec<PolynomialLabel>),
 }
 
+// We implement `PartialEq` manually because its derivation would require
+// `E: PartialEq`. In practice, only `E::G1` and `E::Fr` need it;
+// `E` itself never appears directly in a comparison.
```
davidnevadoc, line 45: Should we specify that this equality is a "literal equality", meaning that equivalent commitments will not return true always. e.g 1) linear commitments whose elements are in a different order. 2) Linear commitments and simple commitments regardless of whether they evaluate to the same commitment or not....

proofs/src/poly/kzg/mod.rs
```diff
@@ -435,22 +416,22 @@ where
         let mut pi_msm = MSMKZG::<E>::init();
         pi_msm.append_term(E::Fr::ONE, pi, PolynomialLabel::Custom("π".into()));
 
-        // Scale zπ - vG
-        let scaled_pi = MSMKZG {
-            scalars: vec![x3, v],
-            bases: vec![pi, -E::G1::generator()],
-            labels: vec![
-                PolynomialLabel::Custom("π".into()),
+        // - vG + zπ
```
davidnevadoc, line 473: Just curious: why the order flip? G first looks better?

proofs/src/poly/kzg/mod.rs
```diff
@@ -435,22 +416,22 @@ where
...
+        let extra_rhs = MSMKZG::new(
+            &[v, x3],
+            &[-E::G1::generator(), pi],
+            &[
                 PolynomialLabel::Custom("-G".into()),
+                PolynomialLabel::Custom("π".into()),
```
davidnevadoc, line 479: Could an issue arise from using weird characters? Probably not but not 100% 🤔

circuits/src/verifier/pcs.rs
```diff
@@ -0,0 +1,149 @@
...
+    ) -> Result<(), Error>;
+
+    /// In-circuit multi-open verification; produces an accumulator.
+    fn multi_prepare(
+        layouter: &mut impl Layouter<S::F>,
+        curve_chip: &S::CurveChip,
+        scalar_chip: &S::ScalarChip,
+        transcript: &mut TranscriptGadget<S>,
+        queries: &[VerifierQuery<'_, S, Self>],
+    ) -> Result<AssignedAccumulator<S>, Error>;
```
davidnevadoc, line 127: Accumulator is the analog of `Guard` in the off-circuit PCS right?
I much prefer this name, so we could consider changing Guard -> Accumulator

circuits/src/verifier/kzg.rs
```diff
@@ -615,3 +529,108 @@ pub(crate) fn multi_prepare<S: SelfEmulation>(
...
+    /// For `Linear`, all existing scalars are multiplied by `scalar`.
+    fn mul(
+        self,
+        layouter: &mut impl Layouter<S::F>,
+        scalar_chip: &S::ScalarChip,
+        scalar: &AssignedNative<S::F>,
+    ) -> Result<Self, Error> {
+        let scalar = AssignedBoundedScalar::new(scalar, None);
+        match self {
+            Self::Simple(p, label) => Ok(Self::Linear(vec![p], vec![scalar], vec![label])),
```
davidnevadoc: 🚀

circuits/src/verifier/pcs.rs
```diff
@@ -0,0 +1,149 @@
...
+        scalar_chip: &S::ScalarChip,
+        other: Self,
+    ) -> Result<Self, Error>;
+}
+
+/// In-circuit abstraction over a Polynomial Commitment Scheme.
+///
+/// Mirrors [`midnight_proofs::poly::commitment::PolynomialCommitmentScheme`]
+/// for the in-circuit verifier.
+pub trait InCircuitPCS<S: SelfEmulation>: Sized + Clone + Debug {
```
davidnevadoc, line 89: The whole commit and overall design is quite clean. I was skeptic about this trait at first, but I like it quite a bit. 
I assume this trait acts  just a sort of API contract, right? We will have just one implementation for it I assume.

proofs/src/poly/kzg/mod.rs
```diff
@@ -491,9 +505,18 @@ mod tests {
     {
         let mut transcript = T::init_from_bytes(proof);
 
-        let a: KZGCommitment<E> = transcript.read().unwrap();
-        let b: KZGCommitment<E> = transcript.read().unwrap();
-        let c: KZGCommitment<E> = transcript.read().unwrap();
+        let a = transcript
```
davidnevadoc, line 545: The increase in lines 😞

proofs/src/poly/kzg/mod.rs
```diff
@@ -68,7 +68,7 @@ where
 {
     type Parameters = ParamsKZG<E>;
     type VerifierParameters = ParamsVerifierKZG<E>;
-    type Commitment = KZGCommitment<E>;
+    type Commitment = KZGMultiCommitment<E>;
```
davidnevadoc, line 71: It can be a bit counter-intuitive for a user that `KZGCommitmentScheme::Commitment` is not `KZGCommitment` when such a structure exists.

Maybe `PolynomialCommitmentScheme` is no longer the adecuate name for this traits, since it is doing something else now. In the same way, `KZGCommitmentScheme` may not be the best name anymore.

davidnevadoc (commented):
Left a couple of comments, but nothing too significante. LGTM! 🚀

.github/workflows/aggregation_examples.yml
miguel-ambrona: You started reviewing this after the base branch got merged into `main` and the history messed up.

proofs/src/poly/kzg/mod.rs
miguel-ambrona: To make it consistent with the in-circuit verifier.

I had to change one of the two. I chose this one for no particular reason. I guess this order makes more sense to me.

proofs/src/poly/kzg/mod.rs
miguel-ambrona: I do not think so in 2026 and under the umbrella of Rust.

circuits/src/verifier/pcs.rs
miguel-ambrona: I agree, but I would do that in a different PR.

circuits/src/verifier/pcs.rs
miguel-ambrona: Yes exactly. There will be just one implementation. But when fflonk comes, this is exactly what fflonk needs to implement.

proofs/src/poly/kzg/mod.rs
miguel-ambrona: Ea.

davidnevadoc (approved):
LGTM!

zk_stdlib/CHANGELOG.md
```diff
@@ -20,7 +20,11 @@ We use [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Fix cost model proof size check to account for committed instance columns [#280](https://github.com/midnightntwrk/midnight-zk/pull/280)
 
 ### Changed
+<<<<<<< HEAD
```
irakoton: Forgotten merge conflict marker

zk_stdlib/src/interface.rs
```diff
@@ -565,15 +566,17 @@ pub fn verify<R: Relation, H: TranscriptHash>(
     params_verifier: &ParamsVerifierKZG<midnight_curves::Bls12>,
     vk: &MidnightVK,
     instance: &R::Instance,
-    committed_instance: Option<G1Affine>,
+    committed_instance: Option<KZGMultiCommitment<midnight_curves::Bls12>>,
```
irakoton, line 569: As far as I understand, this is called multi-commitment now, but only single commitments are supported, and this is the follow-up fflonk branch that implements them.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -198,8 +198,8 @@ impl<S: SelfEmulation> VerifierGadget<S> {
             layouter,
             &self.curve_chip,
             &self.scalar_chip,
-            &[PolynomialLabel::Collapsed],
-            &[fixed_base_labels, &[PolynomialLabel::Collapsed]].concat(),
+            &[PolynomialLabel::NoLabel],
```
irakoton, line 201: _this one is from a review of a previous PR I forgot to submit. Hence the diff not showing up what I'm talking about._

Since the switch to labels, it seems some constraint is missing. The previous version did `assert_equal_to_fixed(scalar, 1)` VS now it just passes the labels and defers to `AssignedAccumulator::assign`, and there's no in-circuit constraint tying the collapsed scalar to 1. Is this normal?

circuits/src/verifier/kzg.rs
```diff
@@ -373,28 +514,41 @@ pub(crate) fn multi_prepare_kzg<S: SelfEmulation>(
     let x1 = transcript_gadget.squeeze_challenge(layouter)?;
     let x2 = transcript_gadget.squeeze_challenge(layouter)?;
 
+    // Peel each query's multi-commitment down to the single inner commitment it
+    // targets, keyed by the query label. A length-1 commitment (the common case,
+    // including the `Linear` linearization commitment) peels to its sole inner;
+    // a batched commitment holds several `Simple`s, so we pick the one whose own
+    // label matches the query.
+    let label_to_commitment: HashMap<PolynomialLabel, &AssignedKZGCommitment<S>> = queries
```
irakoton, line 522: I think it has been solved by now in the string of PRs, but we have to be extra careful, if HashMap / FxHashMap is used as an indexing type, each polynomial should get a different label. I think it hasn't always been the case in the history of our labels.

irakoton (approved):
LGTM, thanks again for the large work. I realised I forgot to hit submit on some previous PRs, and after how they were resolved, one comment I had made was still worth mentioning, so I carried it out there.

Not much to say otheriwse, be simply sure to remove the diff marker. I'll rebase fflonk on top of the corrected, the base branch I was using diverged quite a bit and this will make reviewer's work easier.

zk_stdlib/src/interface.rs
miguel-ambrona: I wouldn't say that only single commitments are supported. This is a structure that represents the commitment to several polynomials.

Internally, it is implemented as a parallel execution of KZG, yes, but that's just an implementation detail. What is important is that the API is "multi" now.

circuits/src/verifier/verifier_gadget.rs
miguel-ambrona: I think I intentionally removed that constrain because it was not necessary.

circuits/src/verifier/kzg.rs
miguel-ambrona: To ensure that, I changed the identifier of some labels from `String` (argument name) to `usize` (argument index).

circuits/src/verifier/verifier_gadget.rs
miguel-ambrona: The fn doc may need to be updated though.

PR #451, opened 2026-06-26
Return a valid Scalar even if it overflows

iquerejeta:
In circuit we allow for Scalar to be represented with more bits than its modulus size, but off-circuit we do not allow that. We need to reduce the bits modulo the scalar field prime order before converting it to `ScalarField`

### DIFF
--- a/circuits/src/ecc/native/edwards_chip.rs
+++ b/circuits/src/ecc/native/edwards_chip.rs
@@ -23,7 +23,7 @@ use midnight_proofs::{
     plonk::{Advice, Column, ConstraintSystem, Constraints, Error, Expression, Selector},
     poly::Rotation,
 };
-use num_bigint::ToBigUint;
+use num_bigint::{BigUint, ToBigUint};
 #[cfg(any(test, feature = "testing"))]
 use {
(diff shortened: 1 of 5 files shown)


davidnevadoc (approved):
LGTM

circuits/src/ecc/native/edwards_chip.rs
```diff
@@ -127,7 +127,15 @@ impl<C: CircuitCurve> InnerValue for AssignedScalarOfNativeCurve<C> {
     fn value(&self) -> Value<Self::Element> {
         let bools = self.bits.iter().map(|b| b.value());
         let value_bools: Value<Vec<bool>> = Value::from_iter(bools);
-        value_bools.map(|le_bits| C::ScalarField::from_bits_le(&le_bits))
+        value_bools.map(|le_bits| {
+            let bytes: Vec<u8> = le_bits
+                .chunks(8)
+                .map(|chunk| chunk.iter().rev().fold(0u8, |acc, &b| (acc << 1) | (b as u8)))
```
miguel-ambrona, line 133: A `reduce` is better than a `fold`  here.

miguel-ambrona (approved):
LGTM

PR #461, opened 2026-07-02
Adapt the in-circuit verifier to have a public input dependent `k`

### LINKED ISSUE #458: Implement the verifier gadget with an assigned `k` and `omega`
### Summary

Currently, the verifier gadget takes `k` and `omega` as fixed, forcing all circuits in the multi-circuit aggregator to have the same size. 

The verifier gadget should allow `k` and `omega` to be treated as public inputs. 

_No response_

### Context / notes

_No response_

iquerejeta:
Closes #458 

This PR treats the domain (concretely `k` and `w`) as proof runtime variables, rather than circuit compile time variables. The regression is 750 rows.

```diff
CircuitModel {
    k: 17,
-   rows: 120830,
+   rows: 121580,
    table_rows: 122899,
    nb_unusable_rows: 11,
    max_deg: 5,
    advice_columns: 15,
    fixed_columns: 49,
    lookups: 2,
    trashcans: 1,
    permutations: 18,
    column_queries: 52,
    point_sets: 5,
    size: 5264,
}
```

### DIFF
--- a/aggregation/examples/multi_circuit_aggregation.rs
+++ b/aggregation/examples/multi_circuit_aggregation.rs
@@ -12,12 +12,15 @@
 //! The requirements for an inner circuit to be aggregatable are:
 //! - It must use the [`ZkStdLibArch`] chosen at IVC setup time.
 //! - It must encode its statement as a single formatted public input.
-//! - It must be padded to the common circuit size `K`.
+//!
+//! Inner circuits may have *different* sizes `K`: the verifier gadget
+//! witnesses each VK's evaluation domain (`k`, `omega`), so a single shared
+//! constraint system (whose shape is independent of `K`) suffices. This
+//! example aggregates a SHA-256 circuit and a Poseidon circuit of different
+//! sizes.
 //!
 //! The single-public-input restriction is not a real limitation: any circuit
-//! can hash its statement into a single field element. On the other hand, the
-//! shared-`K` constraint can be removed by extending the verifier gadget to
-//! accept dynamic domain parameters.
+//! can hash its statement into a single field element.
 //!
 //! DO NOT add this example to the CI as it is slow.
 
@@ -60,14 +63,19 @@ fn main() {
     // Circuit size parameters (log2 of rows).
     const IVC_K: u32 = 19;
 
-    // Shared K for all inner circuits (must accommodate the largest one).
-    const INNER_K: u32 = 13;
+    // Inner circuits may have different sizes.
+    const SHA_K: u32 = 13;
+    const POSEIDON_K: u32 = 14;
 
     // The IVC aggregator only requires a shared SRS and architecture. It does
     // not need to know which circuits will be aggregated. Inner circuits can be
     // introduced, proved and folded in on-the-fly, after IVC initialization.
-    let inner_srs = load_srs(SrsSource::Filecoin, INNER_K, cs_degree(inner_arch()));
-    let inner_ctx = InnerCircuitsContext::new(inner_arch(), INNER_K, inner_srs.verifier_params());
+    // Each circuit is proved against an SRS of its own size.
+    let sha_srs = load_srs(SrsSource::Filecoin, SHA_K, cs_degree(inner_arch()));
+    let poseidon_srs = load_srs(SrsSource::Filecoin, POSEIDON_K, cs_degree(inner_arch()));
+    // Note: verifier params from the SRS do not depend on `k`.
+    let inner_ctx =
+        InnerCircuitsContext::new(inner_arch(), POSEIDON_K, poseidon_srs.verifier_params());
 
     let aggregator_srs = load_srs(
         SrsSource::Midnight,
@@ -79,19 +87,13 @@ fn main() {
     println!("Aggregator setup completed in {:.2?}\n", start.elapsed());
 
     // --- Aggregate a SHA-256 preimage proof (circuit introduced just now) ---
-    let sha_vk = setup_vk(&inner_srs, &ShaCircuit);
+    let sha_vk = setup_vk(&sha_srs, &ShaCircuit);
     let sha_pk = setup_pk(&ShaCircuit, &sha_vk);
 
     let (sha_x, sha_w) = aggregatable_sha_preimage::random_instance();
-    let inner_proof = prove::<ShaCircuit, PoseidonState<F>>(
-        &inner_srs,
-        &sha_pk,
-        &ShaCircuit,
-        &sha_x,
-        sha_w,
-        OsRng,
-    )
-    .expect("SHA-256 proof generation should not fail");
+    let inner_proof =
+        prove::<ShaCircuit, PoseidonState<F>>(&sha_srs, &sha_pk, &ShaCircuit, &sha_x, sha_w, OsRng)
+            .expect("SHA-256 proof generation should not fail");
     let witness = AggregationWitness::new::<ShaCircuit>(sha_vk.clone(), sha_x, inner_proof);
 
     let start = Instant::now();
@@ -99,12 +101,12 @@ fn main() {
     println!("Aggregate a SHA2-256 proof: {:.2?}", start.elapsed());
 
     // --- Aggregate a Poseidon preimage proof (circuit introduced just now) ---
-    let poseidon_vk = setup_vk(&inner_srs, &PoseidonCircuit);
+    let poseidon_vk = setup_vk(&poseidon_srs, &PoseidonCircuit);
     let poseidon_pk = setup_pk(&PoseidonCircuit, &poseidon_vk);
 
     let (poseidon_x, poseidon_w) = aggregatable_poseidon_preimage::random_instance();
     let inner_proof = prove::<PoseidonCircuit, PoseidonState<F>>(
-        &inner_srs,
+        &poseidon_srs,
         &poseidon_pk,
         &PoseidonCircuit,
         &poseidon_x,
@@ -121,15 +123,9 @@ fn main() {
 
     // --- Aggregate another SHA-256 preimage proof ---
     let (sha_x, sha_w) = aggregatable_sha_preimage::random_instance();
-    let inner_proof = prove::<ShaCircuit, PoseidonState<F>>(
-        &inner_srs,
-        &sha_pk,
-        &ShaCircuit,
-        &sha_x,
-        sha_w,
-        OsRng,
-    )
-    .expect("SHA-256 proof generation should not fail");
+    let inner_proof =
+        prove::<ShaCircuit, PoseidonState<F>>(&sha_srs, &sha_pk, &ShaCircuit, &sha_x, sha_w, OsRng)
+            .expect("SHA-256 proof generation should not fail");
     let witness = AggregationWitness::new::<ShaCircuit>(sha_vk, sha_x, inner_proof);
 
     let start = Instant::now();
--- a/aggregation/examples/multi_circuit_aggregation.rs
+++ b/aggregation/examples/multi_circuit_aggregation.rs
@@ -13,9 +13,7 @@
 //! - It must use the [`ZkStdLibArch`] chosen at IVC setup time.
 //! - It must encode its statement as a single formatted public input.
 //!
-//! Inner circuits may have *different* sizes `K`: the verifier gadget
-//! witnesses each VK's evaluation domain (`k`, `omega`), so a single shared
-//! constraint system (whose shape is independent of `K`) suffices. This
+//! Inner circuits may have *different* sizes `K`. Concretely, this
 //! example aggregates a SHA-256 circuit and a Poseidon circuit of different
 //! sizes.
 //!
@@ -64,18 +62,18 @@ fn main() {
     const IVC_K: u32 = 19;
 
     // Inner circuits may have different sizes.
-    const SHA_K: u32 = 13;
-    const POSEIDON_K: u32 = 14;
+    const SHA_CIRCUIT_K: u32 = 13;
+    const POSEIDON_CIRCUIT_K: u32 = 14;
 
     // The IVC aggregator only requires a shared SRS and architecture. It does
     // not need to know which circuits will be aggregated. Inner circuits can be
     // introduced, proved and folded in on-the-fly, after IVC initialization.
     // Each circuit is proved against an SRS of its own size.
-    let sha_srs = load_srs(SrsSource::Filecoin, SHA_K, cs_degree(inner_arch()));
-    let poseidon_srs = load_srs(SrsSource::Filecoin, POSEIDON_K, cs_degree(inner_arch()));
+    let sha_srs = load_srs(SrsSource::Filecoin, SHA_CIRCUIT_K, cs_degree(inner_arch()));
+    let poseidon_srs = load_srs(SrsSource::Filecoin, POSEIDON_CIRCUIT_K, cs_degree(inner_arch()));
     // Note: verifier params from the SRS do not depend on `k`.
     let inner_ctx =
-        InnerCircuitsContext::new(inner_arch(), POSEIDON_K, poseidon_srs.verifier_params());
+        InnerCircuitsContext::new(inner_arch(), POSEIDON_CIRCUIT_K, poseidon_srs.verifier_params());
 
     let aggregator_srs = load_srs(
         SrsSource::Midnight,
--- a/aggregation/examples/multi_circuit_aggregation.rs
+++ b/aggregation/examples/multi_circuit_aggregation.rs
@@ -67,20 +67,19 @@ fn main() {
 
     // The IVC aggregator only requires a shared SRS and architecture. It does
     // not need to know which circuits will be aggregated. Inner circuits can be
-    // introduced, proved and folded in on-the-fly, after IVC initialization.
-    // Each circuit is proved against an SRS of its own size.
-    let sha_srs = load_srs(SrsSource::Filecoin, SHA_CIRCUIT_K, cs_degree(inner_arch()));
-    let poseidon_srs = load_srs(
-        SrsSource::Filecoin,
-        POSEIDON_CIRCUIT_K,
-        cs_degree(inner_arch()),
-    );
+    // introduced, proved and folded on-the-fly, after IVC initialization.
+    // Each circuit is proved against an SRS of its own size whose source (i.e.
+    // toxic waste) is the same.
+    let d = cs_degree(inner_arch());
+
+    let sha_srs = load_srs(SrsSource::Filecoin, SHA_CIRCUIT_K, d);
+    let poseidon_srs = load_srs(SrsSource::Filecoin, POSEIDON_CIRCUIT_K, d);
+
     // Note: verifier params from the SRS do not depend on `k`.
-    let inner_ctx = InnerCircuitsContext::new(
-        inner_arch(),
-        POSEIDON_CIRCUIT_K,
-        poseidon_srs.verifier_params(),
-    );
+    let inner_verifier_params = poseidon_srs.verifier_params();
+
+    let inner_ctx =
+        InnerCircuitsContext::new(inner_arch(), POSEIDON_CIRCUIT_K, inner_verifier_params);
 
     let aggregator_srs = load_srs(
         SrsSource::Midnight,
--- a/aggregation/examples/multi_circuit_aggregation.rs
+++ b/aggregation/examples/multi_circuit_aggregation.rs
@@ -70,10 +70,17 @@ fn main() {
     // introduced, proved and folded in on-the-fly, after IVC initialization.
     // Each circuit is proved against an SRS of its own size.
     let sha_srs = load_srs(SrsSource::Filecoin, SHA_CIRCUIT_K, cs_degree(inner_arch()));
-    let poseidon_srs = load_srs(SrsSource::Filecoin, POSEIDON_CIRCUIT_K, cs_degree(inner_arch()));
+    let poseidon_srs = load_srs(
+        SrsSource::Filecoin,
+        POSEIDON_CIRCUIT_K,
+        cs_degree(inner_arch()),
+    );
     // Note: verifier params from the SRS do not depend on `k`.
-    let inner_ctx =
-        InnerCircuitsContext::new(inner_arch(), POSEIDON_CIRCUIT_K, poseidon_srs.verifier_params());
+    let inner_ctx = InnerCircuitsContext::new(
+        inner_arch(),
+        POSEIDON_CIRCUIT_K,
+        poseidon_srs.verifier_params(),
+    );
 
     let aggregator_srs = load_srs(
         SrsSource::Midnight,
--- a/aggregation/src/ivc/circuit.rs
+++ b/aggregation/src/ivc/circuit.rs
@@ -38,6 +38,11 @@ use super::{Ivc, IvcError, C, F, S};
 #[derive(Clone, Debug)]
 pub struct IvcInstance<T: Ivc> {
     pub(crate) vk_repr: F,
+    // Evaluation domain (`k`, `omega`) of the IVC vk. It is a compile-time
+    // constant, but the circuit emits it as a public input (alongside
+    // `vk_repr`). See [IvcCircuit::format_instance].
+    pub(crate) domain_k: F,
+    pub(crate) domain_omega: F,
     pub(crate) state: T::State,
     pub(crate) acc: Accumulator<S>,
 }
@@ -126,7 +131,7 @@ impl<T: Ivc> Relation for IvcCircuit<T> {
 
     fn format_instance(instance: &Self::Instance) -> Result<Vec<F>, IvcError> {
         Ok([
-            vec![instance.vk_repr],
+            vec![instance.vk_repr, instance.domain_k, instance.domain_omega],
             T::format_public_input(&instance.state),
             AssignedAccumulator::<S>::as_public_input(&instance.acc),
         ]
--- a/aggregation/src/ivc/circuit.rs
+++ b/aggregation/src/ivc/circuit.rs
@@ -145,7 +145,7 @@ impl<T: Ivc> Relation for IvcCircuit<T> {
 
         let assigned_self_vk: AssignedVk<S> = verifier_gadget.assign_vk_as_public_input(
             layouter,
-            &self.domain,
+            Value::known(self.domain.clone()),
             &self.cs,
             instance.as_ref().map(|x| x.vk_repr),
         )?;
--- a/aggregation/src/ivc/circuit.rs
+++ b/aggregation/src/ivc/circuit.rs
@@ -146,12 +146,13 @@ impl<T: Ivc> Relation for IvcCircuit<T> {
         let verifier_gadget = std_lib.verifier();
         let ivc_gadget = T::new(std_lib.clone(), &self.ctx);
 
-        let assigned_self_vk:AssignedVk<S, InCircuitKZG<S>> = verifier_gadget.assign_vk_as_public_input(
-            layouter,
-            &self.cs,
-            Value::known(self.domain.clone()),
-            instance.as_ref().map(|x| x.vk_repr),
-        )?;
+        let assigned_self_vk: AssignedVk<S, InCircuitKZG<S>> = verifier_gadget
+            .assign_vk_as_public_input(
+                layouter,
+                &self.cs,
+                Value::known(self.domain.clone()),
+                instance.as_ref().map(|x| x.vk_repr),
+            )?;
 
         let prev_state_val = witness.as_ref().map(|w| w.prev_state.clone());
         let prev_state = ivc_gadget.assign(layouter, prev_state_val)?;
--- a/aggregation/src/ivc/circuit.rs
+++ b/aggregation/src/ivc/circuit.rs
@@ -129,7 +129,7 @@ impl<T: Ivc> Relation for IvcCircuit<T> {
 
     fn format_instance(instance: &Self::Instance) -> Result<Vec<F>, IvcError> {
         Ok([
-            vec![instance.domain_k, instance.domain_omega, instance.vk_repr],
+            vec![instance.vk_repr, instance.domain_k, instance.domain_omega],
             T::format_public_input(&instance.state),
             AssignedAccumulator::<S>::as_public_input(&instance.acc),
         ]
--- a/aggregation/src/ivc/circuit.rs
+++ b/aggregation/src/ivc/circuit.rs
@@ -39,9 +39,6 @@ use super::{Ivc, IvcError, F, S};
 #[derive(Clone, Debug)]
 pub struct IvcInstance<T: Ivc> {
     pub(crate) vk_repr: F,
-    // Evaluation domain (`k`, `omega`) of the IVC vk. It is a compile-time
-    // constant, but the circuit emits it as a public input (alongside
-    // `vk_repr`). See [IvcCircuit::format_instance].
     pub(crate) domain_k: F,
     pub(crate) domain_omega: F,
     pub(crate) state: T::State,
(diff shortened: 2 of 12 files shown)


circuits/src/verifier/utils.rs
```diff
@@ -140,36 +138,40 @@ pub(crate) fn truncate<F: CircuitField>(
...
-            let x_minus_wi = scalar_chip.add_constant(layouter, x, -wi)?;
+            let wi = if i >= 0 {
+                scalar_chip.pow(layouter, w, i as u64)?
+            } else {
+                scalar_chip.pow(layouter, &w_inv, (-i) as u64)?
+            };
+            let x_minus_wi = scalar_chip.sub(layouter, x, &wi)?;
             let quotient = scalar_chip.div(layouter, &xn_minus_one, &x_minus_wi)?;
-            scalar_chip.mul_by_constant(layouter, &quotient, wi * n_inv)
+            // quotient * w^i / n
```
iquerejeta:
```suggestion
```

circuits/src/verifier/mod.rs
```diff
@@ -52,10 +51,25 @@ pub use verifier_gadget::VerifierGadget;
...
+    n: AssignedNative<S::F>,
+}
+
 /// Type for in-circuit verifying keys.
 ///
-/// This type carries off-circuit a lot of the information about the vk.
-/// The only in-circuit field is the `transcript_repr`.
+/// This type carries off-circuit the information about the constraint system.
+/// The in-circuit fields are the transcript representation, the fixed 
+/// commitments, permutation commitments, and the evaluation domain.
```
miguel-ambrona: This is not completely true.

Even if it says `fixed_commitments: Vec<AssignedKZGCommitment>`, in practice, the assigned KZG commitment is "empty", i.e. of the `Fixed` variant.

I would rephrase this comment, explaining this.

circuits/src/verifier/mod.rs
```diff
@@ -85,7 +99,13 @@ impl<S: SelfEmulation> InnerValue for AssignedVk<S> {
 
 impl<S: SelfEmulation> Instantiable<S::F> for AssignedVk<S> {
     fn as_public_input(vk: &VerifyingKey<S>) -> Vec<S::F> {
-        AssignedNative::<S::F>::as_public_input(&vk.transcript_repr())
+        let domain = vk.get_domain();
+        [
+            AssignedNative::<S::F>::as_public_input(&vk.transcript_repr()),
```
miguel-ambrona: I would place the transcript_repr last.

circuits/src/verifier/mod.rs
```diff
@@ -99,6 +119,16 @@ impl<S: SelfEmulation> AssignedVk<S> {
     pub fn transcript_repr(&self) -> &AssignedNative<S::F> {
         &self.transcript_repr
     }
+
+    /// The assigned `k` cell.
```
miguel-ambrona: cell?

circuits/src/verifier/mod.rs
```diff
@@ -99,6 +119,16 @@ impl<S: SelfEmulation> AssignedVk<S> {
...
+    pub fn k(&self) -> &AssignedNative<S::F> {
+        &self.domain.k
+    }
+
+    /// The assigned `omega` cell.
```
miguel-ambrona: cell?

circuits/src/verifier/utils.rs
```diff
@@ -140,36 +138,40 @@ pub(crate) fn truncate<F: CircuitField>(
...
+    i_indices: Range<i32>,
 ) -> Result<Vec<AssignedNative<F>>, Error> {
-    // For every i, Li(X) := (w^i / n) * (X^n - 1) / (X - w^i).
-
-    let n_inv = F::from(n).invert().unwrap();
-    let xn = scalar_chip.pow(layouter, x, n)?;
-    let xn_minus_one = scalar_chip.add_constant(layouter, &xn, -F::ONE)?;
+    // For every i, L_i(X) := (ωⁱ / n) · (Xⁿ - 1) / (X - ωⁱ).
+    let w_inv = scalar_chip.inv(layouter, w)?;
+    let n_inv = scalar_chip.inv(layouter, n)?;
```
miguel-ambrona: Maybe these could be computed outside and reused. I.e. passed as arguments to evaluate_lagrange_poly.

circuits/src/verifier/utils.rs
```diff
@@ -140,36 +138,40 @@ pub(crate) fn truncate<F: CircuitField>(
...
+    let xn_minus_one = scalar_chip.add_constant(layouter, xn, -F::ONE)?;
 
     i_indices
         .map(|i| {
-            assert!(-(n as i32) <= i);
```
miguel-ambrona, line 167: This assertion went away. I guess it was just a sanity check?

circuits/src/verifier/utils.rs
```diff
@@ -140,36 +138,40 @@ pub(crate) fn truncate<F: CircuitField>(
...
-            let i = if i < 0 { n as i32 + i } else { i };
-            let wi = w.pow([i as u64, 0, 0, 0]);
-            let x_minus_wi = scalar_chip.add_constant(layouter, x, -wi)?;
+            let wi = if i >= 0 {
+                scalar_chip.pow(layouter, w, i as u64)?
+            } else {
+                scalar_chip.pow(layouter, &w_inv, (-i) as u64)?
+            };
```
miguel-ambrona, line 178: Such powers can also be precomputed and shared across all calls to the lagrange poly computation.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -536,7 +540,7 @@ impl<S: SelfEmulation> VerifierGadget<S> {
         // commits to h(X) as a single polynomial (one commitment); otherwise it
         // splits h(X) into `quotient_poly_degree` limbs (one commitment each).
         #[cfg(not(feature = "single-h-commitment"))]
-        let nb_quotient_coms = assigned_vk.domain.get_quotient_poly_degree();
+        let nb_quotient_coms = assigned_vk.cs_degree - 1;
```
miguel-ambrona, line 577: Why this change?

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -995,7 +1003,7 @@ pub(crate) mod tests {
 
     #[derive(Clone, Debug)]
     pub struct TestCircuit {
-        inner_vk: (EvaluationDomain<F>, ConstraintSystem<F>, Value<F>), // (domain, cs, vk_repr)
+        inner_vk: (Value<EvaluationDomain<F>>, ConstraintSystem<F>, Value<F>), // (domain, cs, vk_repr)
```
miguel-ambrona: I would reorder things so that all `Value`s appear together.

aggregation/src/multi_circuit_aggregator/circuit.rs
```diff
@@ -70,6 +70,8 @@ impl State {
 #[derive(Clone, Debug)]
 pub struct AssignedState {
     last_vk_repr: AssignedNative<F>,
+    last_vk_k: AssignedNative<F>,
+    last_vk_omega: AssignedNative<F>,
```
miguel-ambrona: Can we group these 3 in a struct?

aggregation/src/multi_circuit_aggregator/utils.rs
```diff
@@ -46,7 +46,12 @@ pub fn compute_vk_hash(vk: &MidnightVK) -> F {
     let vk = vk.vk();
     let to_raw = Hashable::<PoseidonState<F>>::to_input;
 
-    let vk_repr = vec![vk.transcript_repr()];
+    let domain = vk.get_domain();
+    let vk_repr = vec![
```
miguel-ambrona: This variable should not be called `vk_repr` anymore.

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -12,12 +12,15 @@
 //! The requirements for an inner circuit to be aggregatable are:
 //! - It must use the [`ZkStdLibArch`] chosen at IVC setup time.
 //! - It must encode its statement as a single formatted public input.
-//! - It must be padded to the common circuit size `K`.
+//!
+//! Inner circuits may have *different* sizes `K`: the verifier gadget
+//! witnesses each VK's evaluation domain (`k`, `omega`), so a single shared
+//! constraint system (whose shape is independent of `K`) suffices. This
```
miguel-ambrona: I would remove this comment. It is unclear.

We can simply explain that inner circuits may have different sizes.

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -60,14 +63,19 @@ fn main() {
     // Circuit size parameters (log2 of rows).
     const IVC_K: u32 = 19;
 
-    // Shared K for all inner circuits (must accommodate the largest one).
-    const INNER_K: u32 = 13;
+    // Inner circuits may have different sizes.
+    const SHA_K: u32 = 13;
+    const POSEIDON_K: u32 = 14;
```
miguel-ambrona: Nice.

miguel-ambrona: Maybe `SHA_CIRCUIT_K` and `POSEIDON_CIRCUIT_K` would be even more clear.

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -60,14 +63,19 @@ fn main() {
...
     // not need to know which circuits will be aggregated. Inner circuits can be
     // introduced, proved and folded in on-the-fly, after IVC initialization.
-    let inner_srs = load_srs(SrsSource::Filecoin, INNER_K, cs_degree(inner_arch()));
-    let inner_ctx = InnerCircuitsContext::new(inner_arch(), INNER_K, inner_srs.verifier_params());
+    // Each circuit is proved against an SRS of its own size.
+    let sha_srs = load_srs(SrsSource::Filecoin, SHA_K, cs_degree(inner_arch()));
+    let poseidon_srs = load_srs(SrsSource::Filecoin, POSEIDON_K, cs_degree(inner_arch()));
+    // Note: verifier params from the SRS do not depend on `k`.
+    let inner_ctx =
+        InnerCircuitsContext::new(inner_arch(), POSEIDON_K, poseidon_srs.verifier_params());
```
miguel-ambrona: If they do not depend on k, why do you pass `POSEIDON_K` here?

How about defining a constant called `MAX_K`?

aggregation/src/ivc/circuit.rs
```diff
@@ -38,6 +38,11 @@ use super::{Ivc, IvcError, C, F, S};
 #[derive(Clone, Debug)]
 pub struct IvcInstance<T: Ivc> {
     pub(crate) vk_repr: F,
+    // Evaluation domain (`k`, `omega`) of the IVC vk. It is a compile-time
+    // constant, but the circuit emits it as a public input (alongside
+    // `vk_repr`). See [IvcCircuit::format_instance].
```
miguel-ambrona: Unclear comment. Remove?

aggregation/src/ivc/verifier.rs
```diff
@@ -48,8 +48,15 @@ impl<T: Ivc> IvcVerifier<T> {
...
+        // Reject proofs whose instance claims a different verifying key or a
+        // different evaluation domain than the canonical one. The domain is a
+        // public input, so it must be pinned to the real vk here (it is not
+        // otherwise tied to `vk_repr` in-circuit).
+        let domain = self.vk.vk().get_domain();
+        if instance.vk_repr != self.vk.vk().transcript_repr()
+            || instance.domain_k != F::from(domain.k() as u64)
+            || instance.domain_omega != domain.get_omega()
+        {
             return Err(IvcError::VkMismatch);
```
miguel-ambrona, line 62: This should maybe not called VkMismatch anymore.

Maybe we can introduce a `DomainMismatch`. WDYT?

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -277,6 +248,41 @@ impl<S: SelfEmulation> VerifierGadget<S> {
...
+        let k = self.scalar_chip.assign_fixed(layouter, S::F::from(domain.k() as u64))?;
+        let omega = self.scalar_chip.assign_fixed(layouter, domain.get_omega())?;
+        let domain = self.derive_domain(layouter, k, omega)?;
+        self.assemble_vk(cs, domain, transcript_repr)
+    }
+
+    /// Completes the assigned domain by deriving `omega_inv` and `n = 2^k` from
+    /// assigned `k` and `omega` cells (in-circuit).
+    fn derive_domain(
+        &self,
```
miguel-ambrona, line 287: Clarify that it is the caller's responsibility to ensure the consistency of `k` and `omega`.
Namely, that `omega` is a `2^k`-th root of unity.

Furthermore, not just the consistency but the integrity of these values.
I.e. clarify that these values are trusted at this point and should be constrained properly by the caller.

miguel-ambrona (commented):
Not approving yet as I wanna review it again. These changes are quite delicate.

Could the assigned evaluation domain be part of the assigned vk? That would make a few things more natural.

aggregation/examples/multi_circuit_aggregation.rs
iquerejeta: I'm actually thinking why do we even need to pass in `k` for the context. It's because we define `max_bit_len` to be `k-1`. I'd suggest we change the `InnerCircuitsContext::new` function to take as input `max_bit_len` instead. That's how we define `ZkStdLib::configure`.

aggregation/src/ivc/verifier.rs
iquerejeta: But the verifying key is formed by the domain. IMO the error is clear.

circuits/src/verifier/verifier_gadget.rs
iquerejeta: The domain is no longer available off-circuit (it is assigned now), so `domain.get_quotient_poly_degree()` cannot be called.

circuits/src/verifier/utils.rs
iquerejeta: now the value is assigned, so we cannot do it. I've added a comment saying that it is the caller's responsibility.

circuits/src/verifier/utils.rs
iquerejeta: This is only called twice, one of them being for the instances, which span over negative powers. I don't think it is worth it.

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -60,14 +61,26 @@ fn main() {
...
     // introduced, proved and folded in on-the-fly, after IVC initialization.
-    let inner_srs = load_srs(SrsSource::Filecoin, INNER_K, cs_degree(inner_arch()));
-    let inner_ctx = InnerCircuitsContext::new(inner_arch(), INNER_K, inner_srs.verifier_params());
+    // Each circuit is proved against an SRS of its own size.
+    let sha_srs = load_srs(SrsSource::Filecoin, SHA_CIRCUIT_K, cs_degree(inner_arch()));
+    let poseidon_srs = load_srs(
+        SrsSource::Filecoin,
+        POSEIDON_CIRCUIT_K,
+        cs_degree(inner_arch()),
+    );
```
miguel-ambrona: Assign `cs_degree(inner_arch())` to a variable `d` and it would look nicer.

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -60,14 +61,26 @@ fn main() {
...
+    // Inner circuits may have different sizes.
+    const SHA_CIRCUIT_K: u32 = 13;
+    const POSEIDON_CIRCUIT_K: u32 = 14;
 
     // The IVC aggregator only requires a shared SRS and architecture. It does
     // not need to know which circuits will be aggregated. Inner circuits can be
     // introduced, proved and folded in on-the-fly, after IVC initialization.
-    let inner_srs = load_srs(SrsSource::Filecoin, INNER_K, cs_degree(inner_arch()));
-    let inner_ctx = InnerCircuitsContext::new(inner_arch(), INNER_K, inner_srs.verifier_params());
+    // Each circuit is proved against an SRS of its own size.
```
miguel-ambrona: We should mention that it is important that, even if the SRSs have different sizes, their source is the same (i.e. they have the same toxic waste).

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -60,14 +61,26 @@ fn main() {
...
+    let poseidon_srs = load_srs(
+        SrsSource::Filecoin,
+        POSEIDON_CIRCUIT_K,
+        cs_degree(inner_arch()),
+    );
+    // Note: verifier params from the SRS do not depend on `k`.
+    let inner_ctx = InnerCircuitsContext::new(
+        inner_arch(),
+        POSEIDON_CIRCUIT_K,
+        poseidon_srs.verifier_params(),
```
miguel-ambrona: Give this a generic name in a variable like "inner_verifier_params".

aggregation/src/ivc/circuit.rs
```diff
@@ -147,8 +149,8 @@ impl<T: Ivc> Relation for IvcCircuit<T> {
         let assigned_self_vk: AssignedVk<S, InCircuitKZG<S>> = verifier_gadget
             .assign_vk_as_public_input(
                 layouter,
-                &self.domain,
                 &self.cs,
+                Value::known(self.domain.clone()),
```
miguel-ambrona, line 153: Hum, why do you get the domain from `self` and not from `instance`?

aggregation/src/ivc/verifier.rs
```diff
@@ -47,8 +47,15 @@ impl<T: Ivc> IvcVerifier<T> {
     /// under a different (potentially malicious) circuit could pass
     /// verification.
     pub fn verify(&self, instance: &IvcInstance<T>, proof: &[u8]) -> Result<(), IvcError> {
-        // Reject proofs whose instance claims a different verifying key.
-        if instance.vk_repr != self.vk.vk().transcript_repr() {
+        // Reject proofs whose instance claims a different verifying key or a
+        // different evaluation domain than the canonical one. The domain is a
+        // public input, so it must be pinned to the real vk here (it is not
```
miguel-ambrona: What do you mean by "the domain is a public input"? 

This new comment is unclear to me. Can we rephrase it? The key detail that needs to be explained is the potential mismatch between self and instance.

aggregation/src/multi_circuit_aggregator/circuit.rs
```diff
@@ -183,6 +191,21 @@ impl IvcIO for ProofAggregation {
...
+                    .map(|c| F::from(c.vk.vk().get_domain().k() as u64))
+                    .unwrap_or(F::ZERO)
+            }),
+        )?;
+        let last_vk_omega = self.std_lib.assign(
+            layouter,
+            value.as_ref().map(|s| {
+                s.claims.last().map(|c| c.vk.vk().get_domain().get_omega()).unwrap_or(F::ZERO)
+            }),
+        )?;
```
miguel-ambrona, line 216: There are no consistency nor integrity checks on this k and this omega. I wonder if those are important. Let's add a comment.

aggregation/src/multi_circuit_aggregator/circuit.rs
```diff
@@ -213,18 +242,26 @@ impl IvcIO for ProofAggregation {
...
-        let last_vk_repr =
-            state.claims.last().map(|c| c.vk.vk().transcript_repr()).unwrap_or(F::ZERO);
+        let last_claim = state.claims.last();
+        let last_vk_repr = last_claim.map(|c| c.vk.vk().transcript_repr()).unwrap_or(F::ZERO);
+        let (last_vk_k, last_vk_omega) = last_claim
+            .map(|c| {
+                let d = c.vk.vk().get_domain();
+                (F::from(d.k() as u64), d.get_omega())
+            })
+            .unwrap_or((F::ZERO, F::ZERO));
```
miguel-ambrona: These defaults values for k and omega are clearly not what one would expect. Maybe k=0 and omega=1 would be better. Still, I would add a comment.

aggregation/src/multi_circuit_aggregator/utils.rs
```diff
@@ -46,11 +46,16 @@ pub fn compute_vk_hash(vk: &MidnightVK) -> F {
     let vk = vk.vk();
     let to_raw = Hashable::<PoseidonState<F>>::to_input;
 
-    let vk_repr = vec![vk.transcript_repr()];
+    let domain = vk.get_domain();
+    let in_circuit_vk = vec![
```
miguel-ambrona: why is this called "in-circuit_vk"? I would call it "vk_as_public_inputs" or something like that.

circuits/src/verifier/mod.rs
```diff
@@ -54,10 +53,27 @@ pub use verifier_gadget::VerifierGadget;
 type VerifyingKey<S> =
     plonk::VerifyingKey<<S as SelfEmulation>::F, KZGCommitmentScheme<<S as SelfEmulation>::Engine>>;
 
+/// Type for in-circuit Evaluation Domain.
+///
+/// This type carries only the information needed for the verifier, `k`
+/// and `omega`, and values `omega^{-1}` and `n = 2^k`, computed in-circuit.
+///
+/// The only entry points are via the assignment functions of Verifying Keys.
```
miguel-ambrona, line 61:
```suggestion
/// The only entry points are the assignment functions of Verifying Keys.
```

circuits/src/verifier/mod.rs
```diff
@@ -54,10 +53,27 @@ pub use verifier_gadget::VerifierGadget;
...
+
 /// Type for in-circuit verifying keys.
 ///
-/// This type carries off-circuit a lot of the information about the vk.
-/// The only in-circuit field is the `transcript_repr`.
+/// This type carries off-circuit the information about the constraint system.
+/// The in-circuit fields are the transcript representation and the evaluation
+/// domain. The fixed and permutation commitments are typed as
+/// [AssignedKZGCommitment], but they are "empty", i.e. of the `Fixed` variant:
+/// they only carry a label, no assigned point.
```
miguel-ambrona, line 76: Keep going. Why is that ok? I would add that the actual commitments are provided by the final verifier (the decider).

circuits/src/verifier/mod.rs
```diff
@@ -87,7 +103,13 @@ impl<S: SelfEmulation, PCS: InCircuitPCS<S>> InnerValue for AssignedVk<S, PCS> {
 
 impl<S: SelfEmulation, PCS: InCircuitPCS<S>> Instantiable<S::F> for AssignedVk<S, PCS> {
     fn as_public_input(vk: &VerifyingKey<S>) -> Vec<S::F> {
-        AssignedNative::<S::F>::as_public_input(&vk.transcript_repr())
+        let domain = vk.get_domain();
+        [
+            AssignedNative::<S::F>::as_public_input(&S::F::from(domain.k() as u64)),
+            AssignedNative::<S::F>::as_public_input(&domain.get_omega()),
+            AssignedNative::<S::F>::as_public_input(&vk.transcript_repr()),
```
miguel-ambrona: This order is inconsistent with other paces where "vk_repr" was first, then k and then omega.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -227,45 +233,31 @@ impl<S: SelfEmulation> VerifierGadget<S> {
...
+    /// primitive `2^k`-th root of unity), nor that they are the ones of the
+    /// verifying key identified by `transcript_repr`. It is the caller's
+    /// responsibility to constrain them.
     pub fn assign_vk_as_public_input<PCS: InCircuitPCS<S>>(
         &self,
         layouter: &mut impl Layouter<S::F>,
-        domain: &EvaluationDomain<S::F>,
         cs: &ConstraintSystem<S::F>,
+        domain: Value<EvaluationDomain<S::F>>,
         transcript_repr_value: Value<S::F>,
```
miguel-ambrona, line 254: This function should take `Value<Vk>` instead of `domain` and `transcript_repr_value` separately.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -227,45 +233,31 @@ impl<S: SelfEmulation> VerifierGadget<S> {
...
     ) -> Result<AssignedVk<S, PCS>, Error> {
-        let transcript_repr: AssignedNative<S::F> =
+        let [k, omega] =
```
miguel-ambrona:
```suggestion
        let [k_value, omega_value] =
```

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -274,11 +266,51 @@ impl<S: SelfEmulation> VerifierGadget<S> {
     pub fn assign_fixed_vk<PCS: InCircuitPCS<S>>(
         &self,
         layouter: &mut impl Layouter<S::F>,
-        domain: &EvaluationDomain<S::F>,
         cs: &ConstraintSystem<S::F>,
+        domain: &EvaluationDomain<S::F>,
         transcript_repr_constant: S::F,
```
miguel-ambrona: Can this function take an actual vk instead of the domain and the transcript_repr separately?

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -565,6 +595,13 @@ impl<S: SelfEmulation> VerifierGadget<S> {
         // high probability
         let x = transcript.squeeze_challenge(layouter)?;
 
+        let omega = &assigned_vk.domain.omega;
+        let omega_inv = &assigned_vk.domain.omega_inv;
+        let n = &assigned_vk.domain.n;
```
miguel-ambrona, line 610: Add a comment explaining why `assigned_vk.domain.n` is guaranteed to be bigger than `max_rotation_ and `min_rotation` defined below.

aggregation/src/ivc/circuit.rs
iquerejeta: Does it matter? I'm committing as public input so that equality will be enforced anyways. I'm committing a VK from an `EvaluationDomain`. The instance only contains `omega` and `k`.

circuits/src/verifier/mod.rs
iquerejeta: I was surprised that tests didn't fail, but that's because the inconsistency was both at `as_public_inputs` and `format_instance`.

circuits/src/verifier/verifier_gadget.rs
iquerejeta: Problem is that the VK is assigned as public input from the its own instance. I think its cleaner that way. Otherwise we have to initialise the VK to some value (while the circuit is unknown), and then mutate it to its actual value.

aggregation/examples/multi_circuit_aggregation.rs
```diff
@@ -67,20 +67,19 @@ fn main() {
...
     // Note: verifier params from the SRS do not depend on `k`.
-    let inner_ctx = InnerCircuitsContext::new(
-        inner_arch(),
-        POSEIDON_CIRCUIT_K,
-        poseidon_srs.verifier_params(),
-    );
+    let inner_verifier_params = poseidon_srs.verifier_params();
+
+    let inner_ctx =
+        InnerCircuitsContext::new(inner_arch(), POSEIDON_CIRCUIT_K, inner_verifier_params);
```
miguel-ambrona, line 82: Why passing the `POSEIDON_CIRCUIT_K`? Should we define

```rust
const MAX_INNER_K: usize = max(SHA_CIRCUIT_K, POSEIDON_CIRCUIT_K)
```

aggregation/src/ivc/circuit.rs
miguel-ambrona: What matters to me is the inconsistency between this argument and the next: `instance.as_ref().map(|x| x.vk_repr)`.

Taking it from the instance is cleaner IMHO.
What is really weird is that it is wrapped inside a `Value` if you do it the current way. Because the verifier (at setup) is also passing `Value::known`, so you may as well just remove the `Value` wrapper.

aggregation/src/ivc/verifier.rs
```diff
@@ -47,10 +47,13 @@ impl<T: Ivc> IvcVerifier<T> {
     /// under a different (potentially malicious) circuit could pass
     /// verification.
     pub fn verify(&self, instance: &IvcInstance<T>, proof: &[u8]) -> Result<(), IvcError> {
-        // Reject proofs whose instance claims a different verifying key or a
-        // different evaluation domain than the canonical one. The domain is a
-        // public input, so it must be pinned to the real vk here (it is not
-        // otherwise tied to `vk_repr` in-circuit).
+        // `instance.vk_repr`, `instance.domain_k` and `instance.domain_omega`
+        // are public inputs chosen by the prover: the circuit uses them to
```
miguel-ambrona, line 51: Remove "chosen by the prover".

aggregation/src/ivc/verifier.rs
```diff
@@ -47,10 +47,13 @@ impl<T: Ivc> IvcVerifier<T> {
...
+        // verify the previous step, but nothing forces them to describe the
+        // key this verifier was set up with. Hence we compare all three against
+        // `self.vk` and reject any mismatch; otherwise a proof produced under
+        // a different (possibly malicious) circuit or domain would verify
+        // here.
```
miguel-ambrona, line 56: This whole thing looks weird to me. It was ok-ish before where we had just vk_repr, but now with the domain and all it may make more sense to rename `IvcInstance` to `IvcExtendedInstance` and introduce a new `IvcInstance` without fields `vk_repr`, `domain_k` and `domain_omega`.

That way, `verify` takes only the `state` and the `acc`, clean. No consistency to be checked. And it builds the extended instance before calling `plonk::prepare` from the `self` info.

aggregation/src/multi_circuit_aggregator/circuit.rs
```diff
@@ -252,14 +260,18 @@ impl IvcIO for ProofAggregation {
     }
 
     fn format_public_input(state: &State) -> Vec<F> {
+        // In the base case (no claims yet) there is no last vk, so the values below
+        // are arbitrary placeholders: nothing consumes them, they just need to match
```
miguel-ambrona, line 264: Actually, do they even need to match?

circuits/src/verifier/mod.rs
```diff
@@ -75,11 +75,14 @@ struct AssignedEvaluationDomain<S: SelfEmulation> {
 /// [AssignedKZGCommitment], but they are "empty", i.e. of the `Fixed` variant:
 /// they only carry a label, no assigned point.
 ///
+/// This is fine because the verifier only adds them to an MSM: the accumulator
+/// of [VerifierGadget::prepare] just records their scalars in
+/// `fixed_base_scalars`, and the actual commitments are provided off-circuit by
+/// the final verifier (the decider), via [Accumulator::resolve_fixed_bases].
+/// The key remains bound in-circuit through `transcript_repr`.
+///
```
miguel-ambrona, line 83: Can we rewrite this?

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -236,6 +236,11 @@ impl<S: SelfEmulation> VerifierGadget<S> {
     /// is required off-circuit, except for the `transcript_repr` and the
     /// evaluation domain.
     ///
+    /// These are taken as separate values (instead of a `Value<VerifyingKey>`)
+    /// because a circuit may need to verify its *own* key, in which case no
+    /// verifying key exists yet: the values are then supplied by the prover
+    /// through the public inputs.
+    ///
```
miguel-ambrona, line 243: This comment should disappear if we make my suggestion on `IvcExtendedInstance`.

PR #462, opened 2026-07-02
Improve `VectorGadget` `padding_flag`

### LINKED ISSUE #460: Optimized `padding_flag` in `VectorGadget`
### `padding_flag` does O(M) equality checks where O(M/A + A) suffice

The current implementation is fine, this is just an optimization.

`VectorGadget::padding_flag` (`circuits/src/vec/vector_gadget.rs:131`) builds its front-padding flags with a loop over **every** buffer position `i ∈ [0, M-A]`, doing an `is_equal_to_fixed(start, i)` per position.

But `start` is provably a multiple of `A` , so `is_start` can only fire at chunk boundaries `{0, A, …, M-A}` . Since the flag is constant within a chunk, the boundary bit can just be cloned across the `A` positions (free).

**Fix:** iterate the first loop with `.step_by(A)` and replicate.

davidnevadoc:
# Summary

 - Removes checks that are provably false due to the vector alignment asumptions.
 - Improves test coverage adding last-chunk edge cases.

Closes #460


iquerejeta (dismissed):
Nice improvement!

iquerejeta (approved):

PR #465, opened 2026-07-03
Vector gadget improvements.

davidnevadoc:
Small improvements to prevent misuse of the vector gadget.
 - Add compile-time checks on the invariants that ensure the intended behaviour of the instructions.
 - Add missing tests for `resize()`
 
 - [x] Merge after #462

### DIFF
--- a/circuits/src/vec/vector.rs
+++ b/circuits/src/vec/vector.rs
@@ -35,13 +35,13 @@ use crate::{
 /// sized chunks. The padding at the end of the payload will have a size in
 /// [0, A) such that | front_pad | + | payload | + | back_pad | = M.
 ///
-/// **Invariant: `M` must be a multiple of `A`** (and `A > 0`, `M >= A`).
-/// Several operations (`padding_flag`, `get_limits`, hashing, …) rely on this
-/// to guarantee that the buffer decomposes into exactly `M / A` full chunks
-/// and that a full-capacity vector (`len == M`) can always be placed without
-/// overflow. This is enforced by the entry points
-/// [`assign_with_filler`](crate::instructions::VectorInstructions::assign_with_filler) and
-/// [`assign`](crate::instructions::AssignmentInstructions::assign).
+/// **Invariant: `A > 0`, `A <= M`, and `M` is a multiple of `A`.** Several
+/// operations (`padding_flag`, `get_limits`, hashing, …) rely on this to
+/// guarantee that the buffer decomposes into exactly `M / A` full chunks and
+/// that a full-capacity vector (`len == M`) can always be placed without
+/// overflow.
+/// The invariant is checked at compile time  by [`AssignedVector::new`],
+/// which is the constructor every `AssignedVector` should be built through.
 #[derive(Clone, Debug)]
 pub struct AssignedVector<F: CircuitField, T: Vectorizable, const M: usize, const A: usize> {
     /// Padded payload of the vector. Boxed to keep large buffers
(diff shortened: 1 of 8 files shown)


circuits/src/vec/vector_gadget.rs
```diff
@@ -826,4 +886,21 @@ mod tests {
...
+        let inputs = (0..100).map(|_| F::random(&mut rng)).collect::<Vec<_>>();
+
+        // Grow with different source lengths and alignments.
+        run_resize_vec_test::<_, 64, 2, 128>(&inputs[..40]); // partial, back padding
+        run_resize_vec_test::<_, 64, 2, 128>(&inputs[..64]); // full source vector
+        run_resize_vec_test::<F, 64, 2, 128>(&[]); // empty source vector
+        run_resize_vec_test::<_, 66, 3, 126>(&inputs[..50]); // A = 3
+        run_resize_vec_test::<_, 64, 64, 128>(&inputs[..30]); // single-chunk source (A = M)
+        run_resize_vec_test::<_, 64, 16, 256>(&inputs[..48]); // larger growth
+    }
```
irakoton, line 906: Could we include tests for the failing cases? With `must_panic`, since IIUC they should panic. Typical improper uses of resize would be
1. Resize to `L < M`
2. Resize to `L` that is not a multiple of `A` 
3. IIUC, the value of `A` is shared by `M` and `L` so the invariant `A <= L` should be implied by the rest.
Not sure if L = M should be an issue or not.

circuits/src/vec/vector.rs
```diff
@@ -52,8 +52,31 @@ pub struct AssignedVector<F: CircuitField, T: Vectorizable, const M: usize, cons
     pub(crate) len: AssignedNative<F>,
 }
```
irakoton, line 54: Is there a reason to still keep the internal field `pub(crate)` then? It seems you've carefully replaced the call sites with your new `new` function.

circuits/src/vec/vector.rs
```diff
@@ -52,8 +52,31 @@ pub struct AssignedVector<F: CircuitField, T: Vectorizable, const M: usize, cons
...
+
 /// Returns the range where the data should be placed in the buffer.
 pub fn get_lims<const M: usize, const A: usize>(len: usize) -> Range<usize> {
+    const {
+        assert!(
+            A > 0 && M >= A && M % A == 0,
+            "AssignedVector requires 0 < A <= M and A | M."
+        )
+    };
     let final_pad_len = (A - (len % A)) % A;
```
irakoton, line 95: It's a compile time check so it's not a big deal anyway, but do we really need this assertion here? Since it's enforced when creating an AssignedVector anyway. I'm not against keeping it for safety if deemed preferrable.

irakoton (dismissed):
LGTM! Inlined some comments

circuits/src/vec/vector_gadget.rs
davidnevadoc: Good idea! But we can't really do it. Since the check is performed at compile time, the test fails at compile time, so we cannot get a runtime panic and capture it with `#[should_panic]`

circuits/src/vec/vector.rs
davidnevadoc: I did it mainly to change the code as little as possible, but it is a good idea.
I added a commit where visibility is reduced and a couple of methods are added so that the fields can be safely accessed.

circuits/src/vec/vector.rs
davidnevadoc: It is necessary because this function can be used without an AssignedVector. For example, to make some computation out-of-circuit. So it is good to have the check there as well.

iquerejeta (approved):
LGTM

iquerejeta (approved):
LGTM!

PR #466, opened 2026-07-07
Fflonk PCS

irakoton:
## Changes

Introduces Fflonk as the library's polynomial commitment scheme and folds the previous KZG implementation into it. The diff seems quite big, but don't forget that it involves a lot of KZG code copying into fflonk. I heavily decomposed the diff into small commits to clearly identify the irrelevant diffs.

Fflonk packs up to `2^FFLONK_T_MAX_LOG` polynomials into a single commitment and opens them at the t-th roots of each query point. At `FFLONK_T_MAX_LOG = 0` it degenerates to plain KZG. The only wire-level difference from the previous KZG is a one-byte sub-bundle count prefixed to each commitment group (never hashed), so it is algebraically — but not byte-for-byte — identical, which is why the goldenfiles and static VKs shift by a few bytes.

Some elements (see details in each commit message):

- poly::kzg and poly::fflonk are merged into a single poly::pcs module holding the scheme, the commitment types (FflonkCommitment / FflonkBundle), and the shared primitives kept from KZG (ParamsKZG, MSMKZG, DualMSM, …). The KZGCommitmentScheme / KZGCommitment / KZGMultiCommitment types are removed.
- The PolynomialCommitmentScheme trait is extended for multi-polynomial, SRS-aware schemes (commit_many, srs_monomial_blowup, squeeze_evaluation_point, …), and the PLONK prover/verifier and in-circuit verifier are updated to batch commits and reads per argument. Basically, I added every operations that were needed for fflonk at a protocol level to the PCS trait.
- A temporary poly::kzg compatibility facade (aliases only, no implementation) is kept for the external dependencies like blake2b. It should be removed once that dependency migrates to poly::pcs.

## Caveat

Due to how the proof size is calculated in the cost model in the repo (using `PolynomialCommitmentScheme::commitment_byte_length`), it is currently not taking fflonk batching into account. I.e., even if you change the value of `FFLONK_T_MAX_LOG`, you won't observe a change in the goldenfiles. A small subsequent PR could change the signature of this function to be compatible with fflonk.

## Benchmarks

Results for various parametrisation of fflonk on some examples of zk_stdlib. Benches were temporary (not included in the repo as they were bloating the code base quite a bit). The takeaway is that there tends to be a regression to study. On the prover side, about ~50% of it is due to the bad multi-base management, and will be handle later. Other half is the multi-open argument. Still unclear what's going on on the verifier side.

```
ShaPreImage, K=13

┌───────────┬────────────┬─────────────┬───────────┬──────────────┐
│ T_MAX_LOG │ prove (ms) │ verify (ms) │ proof (B) │ # point sets │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│         0 │        131 │        2.10 │      3472 │            4 │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│         1 │        203 │        2.08 │      3280 │            6 │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│         2 │        237 │        2.13 │      3184 │            5 │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│       3–6 │       ~319 │       ~2.45 │      3136 │            5 │
└───────────┴────────────┴─────────────┴───────────┴──────────────┘

BitcoinThresholdECDSA, K=16

┌───────────┬────────────┬─────────────┬───────────┬──────────────┐
│ T_MAX_LOG │ prove (ms) │ verify (ms) │ proof (B) │ # point sets │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│         0 │       1341 │        2.22 │      3732 │            5 │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│         1 │       1709 │        2.20 │      3540 │            5 │
├───────────┼────────────┼─────────────┼───────────┼──────────────┤
│         2 │       1965 │        2.26 │      3444 │            5 │
└───────────┴────────────┴─────────────┴───────────┴──────────────┘
```


proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,66 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
...
+/// allocations and the sequential reduce chain.
+///
+/// Shared between KZG's and fflonk's `multi_open` (both build the GWC
+/// `f_poly` / `final_poly` this way).
+pub(crate) fn poly_inner_product<F: PrimeField>(
+    polys: &[crate::poly::Polynomial<F, crate::poly::Coeff>],
+    scalars: impl IntoIterator<Item = F>,
+) -> crate::poly::Polynomial<F, crate::poly::Coeff> {
+    use std::marker::PhantomData;
+    let scalars: Vec<F> = scalars.into_iter().take(polys.len()).collect();
```
iquerejeta, line 295: Why not do the collect outside, and take as input a slice of scalars `&[F]`.

proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,66 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
...
+    let max_len = polys.iter().map(|p| p.len()).max().unwrap_or(0);
+    let mut values = vec![F::ZERO; max_len];
+    parallelize(&mut values, |chunk, start| {
+        for (poly, scalar) in polys.iter().zip(scalars.iter()) {
+            let pv: &[F] = poly;
```
iquerejeta, line 300: why this variable?

proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,66 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
...
+            let end = (start + chunk.len()).min(pv.len());
+            if start < pv.len() {
+                for (out, coeff) in chunk[..end - start].iter_mut().zip(&pv[start..end]) {
+                    *out += *coeff * scalar;
+                }
+            }
+        }
+    });
+    crate::poly::Polynomial {
```
iquerejeta, line 309: You should get the full import. You are calling `crate::poly::` a bunch in this function.

proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,66 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
...
+    let mut values = vec![F::ZERO; max_len];
+    parallelize(&mut values, |chunk, start| {
+        for (poly, scalar) in polys.iter().zip(scalars.iter()) {
+            let pv: &[F] = poly;
+            let end = (start + chunk.len()).min(pv.len());
+            if start < pv.len() {
+                for (out, coeff) in chunk[..end - start].iter_mut().zip(&pv[start..end]) {
+                    *out += *coeff * scalar;
+                }
+            }
```
iquerejeta, line 306:
```suggestion
            if let Some(coeffs) = poly.get(start..) {
                for (out, coeff) in chunk.iter_mut().zip(coeffs.iter()) {
                    *out += *coeff * scalar;
                }
            }
```

proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,66 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
...
+    E::G1Affine: CurveAffine<ScalarExt = E::Fr, CurveExt = E::G1>,
+    E::Fr: Ord,
+{
+    let len: usize = msms.iter().map(|m| m.scalars.len()).sum();
+
+    let mut new_scalars = Vec::with_capacity(len);
+    let mut new_bases = Vec::with_capacity(len);
+    let mut new_labels = Vec::with_capacity(len);
+
+    msms.iter_mut().zip(scalars.iter()).for_each(|(msm, s)| {
```
iquerejeta, line 327: With ZIP best to ensure that the size is the size. Given that it is not pub, a debug assertion should suffice.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,467 @@
...
+        match self {
+            Self::Bundle(p, labels) => {
+                assert_eq!(
+                    labels.len(),
+                    1,
+                    "FflonkBundle::Mul: requires t=1; got t={}",
+                    labels.len()
+                );
+                Self::Linear(vec![p], vec![scalar], labels)
+            }
```
iquerejeta, line 163: This is not a fundamental limitation, right? We could eventually add multiplication to arbitrary polynomials.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,467 @@
...
+        let (mut points, mut scalars, mut labels) = match self {
+            Self::Bundle(p, labels) => {
+                assert_eq!(
+                    labels.len(),
+                    1,
+                    "FflonkBundle::Add: requires t=1; got t={}",
+                    labels.len()
+                );
+                (vec![p], vec![E::Fr::ONE], labels)
+            }
```
iquerejeta, line 188: Same here, not fundamental - we could support adding polynomials if the labels have the same size and are equal.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,467 @@
...
+
+/// A fflonk commitment: one or more sub-bundles produced by a single
+/// `commit` call. Length-1 with a single `t=1` `Bundle` is the
+/// linearization-compatible shape.
+///
+/// The const generic `FFLONK_T_MAX_LOG` matches the scheme's: it lets
+/// [`Labelable::label`] reproduce the partition that produced this
+/// commitment, so the `t` of each sub-bundle does not need to be encoded
+/// on the wire. At `FFLONK_T_MAX_LOG = 0` (no bundling) the wire format is
+/// byte-identical to `KZGMultiCommitment`.
```
iquerejeta: Seems the const generic is no longer here - leftover?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,467 @@
...
+#[derive(Clone, Debug)]
+pub struct FflonkCommitment<E: MultiMillerLoop>(pub Vec<FflonkBundle<E>>);
+
+impl<E: MultiMillerLoop> FflonkCommitment<E> {
+    /// A commitment to the zero polynomial: a single singleton sub-bundle at
+    /// the identity point, carrying `label`. Mirrors the former
+    /// `KZGMultiCommitment::commitment_to_zero`.
+    pub fn commitment_to_zero(label: PolynomialLabel) -> Self {
+        Self(vec![FflonkBundle::Bundle(E::G1::identity(), vec![label])])
+    }
```
iquerejeta, line 221: I thought we got rid of this in the other polynomial commitment scheme 🤔  Why is it needed here?

proofs/src/transcript/mod.rs
```diff
@@ -13,8 +13,13 @@ const BLAKE2B_PREFIX_COMMON: u8 = 1;
...
 pub trait TranscriptHash: Clone {
-    /// Input type of the hash function
-    type Input;
+    /// Input type of the hash function.
+    ///
+    /// Bounded `IntoIterator + FromIterator` so that a commitment holding
+    /// several group elements (e.g. an `FflonkCommitment` with multiple
+    /// sub-bundles) can build its transcript input by concatenating the
+    /// per-element inputs.
+    type Input: IntoIterator + FromIterator<<Self::Input as IntoIterator>::Item>;
```
davidnevadoc, line 22: Is there a way to enforce that a certain order is enforced when `Input` is more than one element?

proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,63 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
         .unwrap()
 }
 
+/// Like [`inner_product`] but specialised to coefficient-form polynomials
+/// that may have different lengths (the shorter ones are zero-extended).
```
davidnevadoc, line 285: Just leaving this here as a reminder to check if we actually need different length input. 
If it is not strictly needed, I prefer not supporting it at all.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// via `label`.
+    fn read<R: io::Read>(reader: &mut R, format: SerdeFormat) -> io::Result<Self> {
+        let p = E::G1::read(reader, format)?;
+        Ok(Self::Bundle(p, Vec::new()))
+    }
+
+    fn write<W: io::Write>(&self, writer: &mut W, format: SerdeFormat) -> io::Result<()> {
+        match self {
+            Self::Bundle(p, _) => p.write(writer, format),
+            Self::Linear(..) => panic!("FflonkBundle::Linear cannot be serialized directly"),
```
davidnevadoc, line 111: I am fine with this but just wonder: since we return `io::Result<>`, should we return an Error? Or is that reseved for actual IO errors?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,213 @@
...
+    transcript::{Hashable, TranscriptHash},
+    utils::helpers::{ProcessedSerdeObject, SerdeFormat},
+};
+
+/// A single fflonk sub-bundle: one G1 point committing to `t` polynomials
+/// packed via `combine`, with their labels stored in commit order.
+///
+/// `Linear` is verifier-internal; see the module docs.
+#[derive(Clone, Debug)]
+pub enum FflonkBundle<E: MultiMillerLoop> {
```
davidnevadoc, line 39: On first read, `FflonkBundle` and `Bundle` seem like terms that kind of collide. Maybe we could find better naming. I will make a suggestion if something better comes to mind.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,213 @@
...
+use midnight_curves::pairing::MultiMillerLoop;
+
+use super::FFLONK_T_MAX_LOG;
+use crate::{
+    poly::{commitment::Labelable, query::PolynomialLabel},
+    transcript::{Hashable, TranscriptHash},
+    utils::helpers::{ProcessedSerdeObject, SerdeFormat},
+};
+
+/// A single fflonk sub-bundle: one G1 point committing to `t` polynomials
```
davidnevadoc, line 34: Is there any restriction on `t`? Must it be a power of 2?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+    /// serialized in a single byte (see the `ProcessedSerdeObject` impl), so a
+    /// commitment may hold at most 255 sub-bundles. This is never approached in
+    /// practice: the count is `ceil(polys_in_group / 2^FFLONK_T_MAX_LOG)`,
+    /// which is maximal at `FFLONK_T_MAX_LOG = 0` and stays in the low
+    /// double digits for real circuits. Panics rather than silently
+    /// truncating if exceeded.
+    fn sub_bundle_count_u8(&self) -> u8 {
+        u8::try_from(self.0.len()).expect(
+            "FflonkCommitment holds more than 255 sub-bundles; the wire format \
+             uses a single-byte count. Widen the prefix (e.g. LEB128) to support this.",
```
davidnevadoc: This is never going to be triggered. But if it was, I believe the suggestion may not be enough as the assumption is baked in other places. For example, just a few lines below in this file:
 https://github.com/midnightntwrk/midnight-zk/pull/466/changes/bc74738bf87d53550ad3897a0c62c7a2a0a53efc#diff-0f449dd54ea2e8aaea8a87327c3f10b2b2ba5bac5ead8562d7d6c5e6fdb4488bR372

Edit: Just saw it in read functions as well, it is hardcoded in many places.

So I'd remove the suggestion and just say that 255 is the max supported number of sub-bundles.

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//!   first-occurrence order. Queries are emitted in the chain order of
+//!   `compute_queries`, which doesn't necessarily match the canonical-key sort
+//!   within any one family.
+//!
+//! If `partition` preserved input order, the two callers would build
+//! different `g(X) = Σ_i X^i f_i(X^t)` polynomials (different slot ↔ label
+//! mapping), the prover's commitment and the verifier's reconstruction
+//! would disagree, and the proof would fail to verify on honest inputs.
+//!
+//! The fix: `partition` sorts each bundlable label family by a canonical
```
davidnevadoc: 🚀

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//! `partition` reordered across families.
+//!
+//! Resolved: `multi_open` now walks `queries` directly (in their natural
+//! emission order) to build `singleton_queries`, mirroring the verifier's
+//! `bundle_expansion::classify_verifier_queries`. The partition output is
+//! consulted only to identify which labels live in a `t > 1` bundle (so
+//! those queries flow through the bundle-expansion path instead). The
+//! cross-family emission order in the partition output is therefore no
+//! longer load-bearing — it remains deterministic for testability but is
+//! a free parameter for adding new families.
```
davidnevadoc: This is a nice bit of history but I think we could summarize it with an explanation of the end result only 😁

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//! lives at keygen time and is a separate workstream (it changes the VK
+//! shape).
+//!
+//! ## `LogupHelper` labels
+//!
+//! `LogupHelper(name, chunk)` carries both the argument name and the
+//! chunk index — every helper polynomial has a unique label, so cross-arg
+//! and within-arg helper bundling are both well-defined.
+//!
+//! ## Sparsity caveat for `LagrangeDelta`-basis families
```
davidnevadoc: Let's mark this as a TODO so we don't forget removing it when we optimize for sparsity.

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//! when populating the `Option<usize>` field). A multi-poly fflonk bundle
+//! exposes only the combined group element
+//! `[g(τ)]_1 = Σ_k τ^k h_k(τ^t)`, from which the individual `[h_k(τ)]_1`s
+//! cannot be recovered. Bundling these without first rewriting linearization
+//! would break soundness.
+//!
+//! Non-selector fixed columns (lookup tables, generic precomputed columns)
+//! are *not* touched by linearization — they're opened through
+//! `compute_queries`'s `fixed_queries` chain and could be bundled. That
+//! lives at keygen time and is a separate workstream (it changes the VK
```
davidnevadoc: Let's eliminate the workstream reference

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//! - Bundlable families (`Advice`, `CommittedInstance`,
+//!   `PermutationAccumulator`, `LogupHelper`, `LogupMultiplicities`,
+//!   `LogupAggregator`) are grouped per-family, sorted by their family's
+//!   canonical key (`usize` index for the numeric variants, lexicographic
+//!   `String` for the named `Logup*` variants), and chunked into sub-bundles of
+//!   size `≤ t_max`.
+//! - Non-bundlable labels are committed as their own singleton sub-bundles (`t
+//!   = 1`).
+//!
+//! ## Why `QuotientPiece` and simple-selector `Fixed` stay singletons
```
davidnevadoc: Are we keeping quotient pieces? If the SRS now is large enough, don't we commit to the whole polynomial in one go?

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+        PolynomialLabel::LogupHelper(..) => Some(BundleFamily::LogupHelper),
+        PolynomialLabel::LogupMultiplicities(_) => Some(BundleFamily::LogupMultiplicities),
+        PolynomialLabel::LogupAggregator(_) => Some(BundleFamily::LogupAggregator),
+        _ => None,
+    }
+}
+
+/// Sort a bin of label-indices by the family's canonical key. Load-bearing
+/// — see module docs. Mixing variants within one bin is impossible by
+/// construction (each bin is one family).
```
davidnevadoc: Fine, just making a note. There may be a more performant way of ensuring order is not a problem and maybe with less code.

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+
+/// Sub-bundle partition of `labels`. Returns a list of index-vecs into
+/// `labels`; each inner vec is one sub-bundle, with indices in canonical
+/// order shared between prover and verifier.
+///
+/// Output order: bundlable families first (each sorted by its canonical
+/// key, with families themselves ordered by [`BundleFamily`]'s `Ord`),
+/// then singletons in input order.
+///
+/// # Soundness for linearised polynomials
```
davidnevadoc, line 118: Is there an analysis on the interaction of fflonk + quotient split + linearization?
Is the best option to combine them all? They seem to be in conflict.

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+    }
+    result
+}
+
+/// Logical bundle size for a sub-bundle with `real_count` real polynomials
+/// at scheme cap `t_max`. This is because fflonk requires bundle's sizes to be
+/// a power of two, which makes it necessary to pad bundles with null
+/// polynomials until the next power.
+pub(super) fn bundle_t(real_count: usize, t_max: usize) -> usize {
+    real_count.next_power_of_two().min(t_max)
```
davidnevadoc, line 153: `t_max` is a variable in this and other functions but it is known at compile time. Why not make it a constant?

davidnevadoc (commented):
First pass of the 8 first commits

proofs/src/utils/arithmetic.rs
```diff
@@ -279,6 +281,63 @@ pub(crate) fn inner_product<F: PrimeField, T: Mul<F, Output = T> + Add<T, Output
...
+///
+/// Fused parallel implementation: a single pass accumulates all scaled
+/// contributions directly into the output buffer, avoiding `M` intermediate
+/// allocations and the sequential reduce chain.
+pub(crate) fn poly_inner_product<F: PrimeField>(
+    polys: &[crate::poly::Polynomial<F, crate::poly::Coeff>],
+    scalars: impl IntoIterator<Item = F>,
+) -> crate::poly::Polynomial<F, crate::poly::Coeff> {
```
iquerejeta, line 293: Is it intentional that we don't enforce that `polys` and `scalars` have the same `len`?

proofs/src/poly/commitment.rs
```diff
@@ -68,6 +68,48 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
+    /// verifier can compute `t`-th roots when reconstructing `g(root)` from
+    /// per-slot evaluations (Lemma 5.1).
+    ///
+    /// Protocol callers (both prover and verifier) should obtain the
+    /// evaluation point through this method rather than calling
+    /// `transcript.squeeze_challenge()` directly, so swapping in a
+    /// different PCS keeps the protocol generic. No `Parameters` argument
+    /// is taken so the method is symmetric across prover and verifier;
+    /// schemes that need to vary the transform based on SRS sizing should
+    /// encode it at the type level (e.g. as a const generic).
```
iquerejeta: Note to self: I'm not fully convinced about having this here. I'll get back to this at the end of the review.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+
+    fn read(buffer: &mut impl Read) -> io::Result<Self> {
+        let mut len_bytes = [0u8; 1];
+        buffer.read_exact(&mut len_bytes)?;
+        let len = len_bytes[0] as usize;
+        let inner = (0..len)
+            .map(|_| <FflonkBundle<E> as Hashable<H>>::read(buffer))
+            .collect::<io::Result<Vec<_>>>()?;
+        Ok(Self(inner))
+    }
```
iquerejeta, line 382: Given that you've implemented `ProcessedSerdeObject` for `FflonkCommitment`, you can use those functions here.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+}
+
+impl<E: MultiMillerLoop> Mul<E::Fr> for FflonkCommitment<E> {
+    type Output = Self;
+
+    fn mul(self, scalar: E::Fr) -> Self {
+        self.assert_single_sub_bundle();
+        Self(vec![self.0.into_iter().next().unwrap() * scalar])
```
iquerejeta, line 390:
```suggestion
        Self(vec![self.0[0] * scalar])
```

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+}
+
+impl<E: MultiMillerLoop> Add for FflonkCommitment<E> {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        self.assert_single_sub_bundle();
+        other.assert_single_sub_bundle();
+        Self(vec![
+            self.0.into_iter().next().unwrap() + other.0.into_iter().next().unwrap(),
```
iquerejeta, line 401:
```suggestion
            self.0[0] + other.0[0], 
```

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+/// `bundle_labels` arrives in `partition`(super::partition::partition)'s
+/// canonical order, so `bundle_labels[0]` is the family's canonical-first
+/// label. Using it (rather than scanning for a fixed family like `Advice`)
+/// generalises across every bundlable family: Advice bundles produce
+/// `fflonk_g[advice_<min>]`, PermutationAccumulator bundles produce
+/// `fflonk_g[perm_acc_<min>]`, and so on — all derived uniformly from the
+/// `Display` impl of [`PolynomialLabel`].
+pub(super) fn synthetic_bundle_label(bundle_labels: &[PolynomialLabel]) -> PolynomialLabel {
+    let first = bundle_labels.first().expect("fflonk: multi-poly bundle must be non-empty");
+    PolynomialLabel::Custom(format!("fflonk_g[{}]", first))
```
iquerejeta, line 410: What's the `_g` for? What about using `fflonk_bundle[{}]` instead?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+///
+/// Match policy:
+/// - [`FflonkBundle::Bundle`] matches if `label` is among its own labels (the
+///   standard case for advice / instance / fixed / quotient commitments).
+/// - [`FflonkBundle::Linear`] matches unconditionally — by construction,
+///   `Linear` only ever appears as a linearization commitment, whose query
+///   carries the synthetic [`PolynomialLabel::NoLabel`] rather than any of the
+///   inner component labels. This mirrors how KZG's `multi_prepare` treats
+///   `KZGCommitment::Linear`.
+pub(super) fn find_bundle<'a, E: MultiMillerLoop>(
```
iquerejeta, line 424: why not an associated function of `FflonkCommitment`?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -211,3 +211,269 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+                    FflonkBundle::Bundle(_, ls) => ls.clone(),
+                    FflonkBundle::Linear(_, _, ls) => ls.clone(),
+                })
+                .collect();
+            panic!(
+                "fflonk multi_prepare: query label {:?} not found in any bundle of its commitment; \
+                 bundle labels were {:?}",
+                label, bundle_labels,
+            )
+        })
```
iquerejeta, line 449: Too much code for error handling. I'd just panic with text "query label not found in any bundle of its commitment".

proofs/src/poly/pcs/partition.rs
iquerejeta: No need to explain how it used to be implemented, and the issues we had during development.

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//! `T_MAX = 2^FFLONK_T_MAX_LOG` is the scheme-wide const generic of
+//! [`FflonkScheme`](super::FflonkScheme), so both sides agree on `t_max`
+//! without consulting the SRS. Any change to the partition strategy is a
+//! *protocol-level* change (commitments produced differ) and must be
+//! versioned together with VK updates.
+//!
+//! ## Bundling policy
+//! - Bundlable families (`Advice`, `CommittedInstance`,
+//!   `PermutationAccumulator`, `LogupHelper`, `LogupMultiplicities`,
+//!   `LogupAggregator`) are grouped per-family, sorted by their family's
```
iquerejeta, line 13: trash argument?

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+//! 2. Rewrite linearization to consume openings rather than per-poly
+//!    commitments, unblocking `QuotientPiece` and simple-selector `Fixed`
+//!    bundling.
+//! 3. Split `vk.fixed_commitments` into simple-selector vs. non-selector
+//!    regions and bundle the non-selector half at keygen.
+//! 4. Phase-merging: collapse phases whose commits don't depend on a
+//!    later-squeezed challenge into one batched commit (paper §2.3 enables
+//!    heterogeneous bundles across bases).
+//! 5. Per-phase variable `t` (decoupling sub-bundle sizes from the
+//!    protocol-level `T_MAX`).
```
iquerejeta: The docs should focus on what's implemented. I'd suggest that follow ups are tracked as issues, and not doc comments.

proofs/src/poly/pcs/partition.rs
iquerejeta: the linter added a new line in a odd spot. Put 'Load-bearing - see module docs' in the same line.

Why is this the case, though? I think we should precisely allow mixing variants.

proofs/src/poly/pcs/partition.rs
```diff
@@ -0,0 +1,369 @@
...
+/// module docs).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(super) enum BundleFamily {
+    Advice,
+    CommittedInstance,
+    PermutationAccumulator,
+    LogupHelper,
+    LogupMultiplicities,
+    LogupAggregator,
+}
```
iquerejeta, line 38: I would really try to avoid this abstraction, and allow different polynomial labels to be bundled together. What is the limitation?

proofs/src/poly/pcs/bundle_expansion.rs
```diff
@@ -0,0 +1,288 @@
+//! Bundle pre-expansion data structures and pure helpers.
+//!
+//! `FflonkScheme::multi_open` and `FflonkScheme::multi_prepare` each have a
+//! "fflonk-specific" phase that runs before the standard Halo2 GWC argument:
+//!
+//! - Prover (`multi_open`): classify queries into bundles, materialise the
+//!   combined polynomial `g` for each `t > 1` bundle, compute the over-opening
```
iquerejeta, line 7: what is `over opening`?

proofs/src/poly/pcs/bundle_expansion.rs
```diff
@@ -0,0 +1,288 @@
...
+    utils::{
+        arithmetic::{CurveAffine, CurveExt, MSM},
+        helpers::ProcessedSerdeObject,
+    },
+};
+
+/// Per-bundle prover-side preparation. Holds everything `multi_open` needs
+/// to (a) write over-opening evals to the transcript and (b) emit the
+/// synthetic queries on `g`.
+pub(super) struct MultiPreData<F> {
```
iquerejeta, line 39: name suggestion: `MultiOpenPrepData`

iquerejeta (commented):
Haven't started reviewing the PCS.

proofs/src/poly/pcs/partition.rs
irakoton: A hell of a leftover from a session with friend Claudio 😮‍💨 I had made a big pass to clean up the doc, but it seems I forgot to pass on partition.rs. I made another pass on the whole repo and there were a few isolated places with bloated comments I had missed.

For this one, I will force push so that it doesn't interfere with the later reviews. I'll address your technical comments with new commits.

proofs/src/poly/pcs/partition.rs
irakoton: Should I remove all mentions to linearisation? I feel otherwise it's a bit hard to understand why some label families are left as singletons.

proofs/src/poly/pcs/partition.rs
irakoton: Yeah that's my bad. This crate level doc was generated by Claude, and I visibly forgot to make a cleaning pass on the comments of partition.rs. I made a check on the whole repo after your and David's remark, this should now be good.

proofs/src/poly/pcs/msm.rs
miguel-ambrona, line 1: At this point I think it makes sense to have a `src/poly` with just the poly arithmetic and then `src/pcs` or even more explicit `src/polynomial_commitment`.

proofs/src/poly/commitment.rs
```diff
@@ -68,6 +68,33 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
+    /// verifying certain properties, for example fflonk requires the
+    /// evaluation point to be a `t`-th power in the field.
+    ///
+    /// The protocol must squeeze evaluation points through this method.
+    fn squeeze_evaluation_point<T: Transcript>(transcript: &mut T) -> F
+    where
+        F: Sampleable<T::Hash>,
+    {
+        transcript.squeeze_challenge()
+    }
```
miguel-ambrona, line 83: This makes a lot of sense. 👏

proofs/src/poly/commitment.rs
```diff
@@ -68,6 +68,33 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
+    /// commit every polynomial it produces at the requested circuit size.
+    /// Returns `1` when no extension is needed.
+    ///
+    /// `cs_degree` is the constraint system's `cs.degree()`. Schemes that
+    /// commit to a single combined polynomial (e.g. `single-h-commitment`,
+    /// fflonk's bundles) factor that into their requested blow-up.
+    fn srs_monomial_blowup(cs_degree: usize) -> usize {
+        let _ = cs_degree; // Just to avoid a clippy warning.
+        1
+    }
```
miguel-ambrona, line 96: I'm not fond of this function.

The fact that the srs is structured in monomials should be an implementation detail.

Maybe it makes more sense to remove this function and modify `gen_params` to take `max_poly_degree` instead of `k` (which btw is weird). And the "blowup" detail will appear in the implementation of `gen_params`.

proofs/src/poly/commitment.rs
```diff
@@ -89,8 +116,15 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
 
     /// Verify an multi-opening proof for a given set of [VerifierQuery]'s.
     /// The function fails if the transcript has trailing bytes.
+    ///
+    /// `k` is the log2 of the circuit's (Lagrange) domain size. Schemes whose
```
miguel-ambrona, line 120: We should not be talking about "circuits" in this trait.

proofs/src/poly/commitment.rs
```diff
@@ -89,8 +116,15 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
+    /// opening structure depends on the domain relative to the SRS capacity
+    /// (e.g. fflonk's SRS-aware bundling) need it to reconstruct and
+    /// sanity-check the prover's choices; schemes that don't (e.g. plain KZG)
+    /// ignore it.
     fn multi_prepare<'com, T: Transcript>(
         verifier_query: &[VerifierQuery<'com, F, Self>],
+        k: u32,
```
miguel-ambrona, line 127: If you need `k`, it may make more sense to require the verifier params.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
+//! fflonk commitment types.
+//!
+//! A [`FflonkCommitment`] is the output of one `commit` call: a `Vec` of
+//! [`FflonkBundle`]s, one per sub-bundle produced by `partition`. Each
```
miguel-ambrona, line 4: sub-bundle or bundle?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+//!
+//! [`FflonkBundle::Linear`] is a verifier-internal deferred-MSM accumulator
+//! produced by `Add`/`Mul<F>` on `t=1` sub-bundles, used by linearization.
+//! It is never serialized or hashed (panic on attempt).
+//!
+//! # `Add` / `Mul<F>` semantics
+//! These trait-level bounds on `Commitment` exist for the linearization MSM
+//! (`proofs/src/plonk/linearization/verifier.rs`), which operates on
+//! single-polynomial commitments. Attempting `Add`/`Mul` on a sub-bundle
+//! with `t > 1` panics.
```
miguel-ambrona, line 16: All this should belong in a fflonk folder, not here.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+}
+
+impl<H: TranscriptHash, E: MultiMillerLoop> Hashable<H> for FflonkCommitment<E>
+where
+    E::G1: Hashable<H> + Default + ProcessedSerdeObject,
+{
+    fn to_input(&self) -> H::Input {
+        self.0.iter().flat_map(|b| b.to_input()).collect()
+    }
```
remalue, line 364: Does this include the `sub_bundle_count_u8`? It should be hashed as well, no?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    /// Returns a reference to the inner curve point, panicking on `Linear`.
+    pub fn as_point(&self) -> &E::G1 {
+        match self {
+            Self::Bundle(p, _) => p,
+            Self::Linear(..) => panic!("expected FflonkBundle::Bundle"),
+        }
+    }
```
miguel-ambrona, line 62: I would pick one: either `into_point` or `as_point` but why both? You probably want the one that deals with references.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+// This comparison is literal. Two equivalent commitments are considered
+// different if their representation is different.
+impl<E: MultiMillerLoop> PartialEq for FflonkBundle<E>
+where
+    E::G1: PartialEq,
+    E::Fr: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Bundle(p, l1), Self::Bundle(q, l2)) => p == q && l1.len() == l2.len(),
```
miguel-ambrona, line 78: You lied, you are only comparing the length of the labels, not the actual labels.

miguel-ambrona: In `Linear` you do not even compare the labels lengths. Why? 😢

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+impl<E: MultiMillerLoop> Default for FflonkBundle<E>
+where
+    E::G1: Default,
+{
+    fn default() -> Self {
+        Self::Bundle(E::G1::default(), vec![PolynomialLabel::NoLabel])
+    }
+}
```
miguel-ambrona, line 92: I would avoid implementing `Default` if possible.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+
+impl<E: MultiMillerLoop> ProcessedSerdeObject for FflonkBundle<E>
+where
+    E::G1: Default + ProcessedSerdeObject,
+{
+    /// Wire format: a single G1 point. The bundle's `t` (number of
+    /// polynomials packed into this sub-bundle) is not encoded as it is
```
miguel-ambrona, line 99: bundle or sub-bundle?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// recovered from `partition`(super::partition::partition). Labels are an
+    /// empty placeholder after deserialization and must be attached via
+    /// `label`.
+    fn read<R: io::Read>(reader: &mut R, format: SerdeFormat) -> io::Result<Self> {
```
miguel-ambrona, line 103: I do not remember if we actually did it in the end, but I believe the plan was to provide `t` (or even the labels) to the `read` function. We can polish this in a future PR if necessary.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    fn to_bytes(&self) -> Vec<u8> {
+        match self {
+            Self::Bundle(p, _) => p.to_bytes(),
+            Self::Linear(..) => panic!("FflonkBundle::Linear cannot be serialized to bytes"),
+        }
+    }
+
+    fn read(buffer: &mut impl Read) -> io::Result<Self> {
```
miguel-ambrona, line 141: Another `read`? Why?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+            Self::Bundle(p, labels) => {
+                assert_eq!(
+                    labels.len(),
+                    1,
+                    "FflonkBundle::Add: requires t=1; got t={}",
+                    labels.len()
+                );
+                (vec![p], vec![E::Fr::ONE], labels)
+            }
+            Self::Linear(points, scalars, labels) => (points, scalars, labels),
```
miguel-ambrona, line 189: You could have a helper function to do this match and use it in 3 places (including the impl of `mul`).

proofs/src/poly/pcs/commitment.rs
```diff
@@ -207,3 +207,244 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+impl<E: MultiMillerLoop> FflonkCommitment<E> {
+    /// A commitment to the zero polynomial: a single singleton sub-bundle at
+    /// the identity point, carrying `label`.
+    pub fn commitment_to_zero(label: PolynomialLabel) -> Self {
+        Self(vec![FflonkBundle::Bundle(E::G1::identity(), vec![label])])
+    }
+
+    /// Sub-bundle count as the `u8` written on the wire. The count is
+    /// serialized in a single byte, so a commitment may hold at most 255
+    /// sub-bundles.
```
miguel-ambrona, line 225: Well, it would be better to document the length limit in the actual `commit` function too or even the commitment type.

I would try to relax the limit if possible. We can use 4 bytes for the length or use some kind of "prefix code" to use 1 byte in most cases. For example, if the first bit is 0, the next 7 bits encode a length in the range [0, 128).
Otherwise, the next 7 bits followed by the next 3 bytes encode a length. This would make the proof length variable depending on the CS architecture, but I'm afraid it is already variable w.r.t. the CS, so I think we are good.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -207,3 +207,244 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
     }
 }
 
+/// A fflonk commitment: one or more sub-bundles produced by a single
+/// `commit` call. Length-1 with a single `t=1` `Bundle` is the
+/// linearization-compatible shape.
+#[derive(Clone, Debug)]
+pub struct FflonkCommitment<E: MultiMillerLoop>(pub Vec<FflonkBundle<E>>);
```
miguel-ambrona, line 214: I'm not sure we need both `FflonkCommitment` and `FflonkBundle`.

Couldn't we just have this?
```rust
enum FflonkCommitment<E> {
  Normal(Vec<E::G1>, Vec<PolynomialLabel>),
  Linear(Vec<E::G1>, Vec<PolynomialLabel>),
}
```

This works because we only allow to combine "linearly" things with `t = 1`.

miguel-ambrona: You may rename "Normal" if you go for this. I don't have a good suggestion ATM.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -207,3 +207,244 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+}
+
+impl<E: MultiMillerLoop> Default for FflonkCommitment<E>
+where
+    E::G1: Default,
+{
+    fn default() -> Self {
+        Self(vec![FflonkBundle::default()])
+    }
+}
```
miguel-ambrona, line 272: Remove `Default` if possible.

proofs/src/poly/pcs/utils.rs
```diff
@@ -4,16 +4,110 @@ use std::{
...
+    values
+}
+
+/// Returns a primitive `t`-th root of unity in `F`, where `t` is a power of
+/// two not exceeding the field's 2-adicity `F::S`. Computed as
+/// `F::ROOT_OF_UNITY^(2^(S − log2 t))`.
+///
+/// # Panics
+/// If `t` is not a power of two, or `log2 t > F::S`.
+pub(super) fn primitive_root_of_unity<F: PrimeField>(t: usize) -> F {
```
miguel-ambrona, line 50: Doesn't this function exist somewhere else?

proofs/src/poly/pcs/mod.rs
```diff
@@ -233,3 +233,131 @@ where
...
+                                    let dd_poly: Polynomial<E::Fr, LagrangeDoubleDeltaCoeff> =
+                                        Polynomial {
+                                            values: polynomials[i].values.clone(),
+                                            _marker: PhantomData,
+                                        };
+                                    let lagrange = dd_poly.into_lagrange();
+                                    domain.lagrange_to_coeff(lagrange).values
+                                })
+                                .collect()
+                        }
```
miguel-ambrona, line 341: This code is really ugly.

proofs/src/poly/commitment.rs
```diff
@@ -68,6 +68,33 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
+    /// Multiplicative blow-up factor by which `params.g_monomial_size()` must
+    /// exceed `2^k` (the circuit's Lagrange-domain size) for this PCS to
+    /// commit every polynomial it produces at the requested circuit size.
+    /// Returns `1` when no extension is needed.
+    ///
+    /// `cs_degree` is the constraint system's `cs.degree()`. Schemes that
+    /// commit to a single combined polynomial (e.g. `single-h-commitment`,
+    /// fflonk's bundles) factor that into their requested blow-up.
+    fn srs_monomial_blowup(cs_degree: usize) -> usize {
+        let _ = cs_degree; // Just to avoid a clippy warning.
```
remalue, line 94: Maybe `#[allow(unused_variables)]`?

proofs/src/poly/commitment.rs
```diff
@@ -89,8 +116,15 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
 
     /// Verify an multi-opening proof for a given set of [VerifierQuery]'s.
     /// The function fails if the transcript has trailing bytes.
+    ///
+    /// `k` is the log2 of the circuit's (Lagrange) domain size. Schemes whose
+    /// opening structure depends on the domain relative to the SRS capacity
+    /// (e.g. fflonk's SRS-aware bundling) need it to reconstruct and
+    /// sanity-check the prover's choices; schemes that don't (e.g. plain KZG)
```
remalue, line 123: What sanity checks/choices are you referring to?

miguel-ambrona (commented):
Leaving a few comments.

I plan to make a more serious pass on the Fflonk verifier.

proofs/src/poly/pcs/commitment.rs
miguel-ambrona: Actually, we may not need to serialize the size if we know how to parse it when reading.

proofs/src/poly/pcs/commitment.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// A single fflonk sub-bundle: one G1 point committing to `t` polynomials
+/// packed via `combine`, with their labels stored in commit order.
+///
+/// `Linear` is verifier-internal; see the module docs.
+#[derive(Clone, Debug)]
+pub enum FflonkBundle<E: MultiMillerLoop> {
+    /// One G1 commitment to `t = labels.len()` polynomials. `t` may be 1.
+    Bundle(E::G1, Vec<PolynomialLabel>),
+    /// Lazy linear combination `\sum scalars[i] * points[i]` with per-term
```
remalue, line 42: I like the term "lazy linear combination" but stumbled over it when I read it the first time. Maybe define it somewhere appropriate?

proofs/src/poly/pcs/commitment.rs
```diff
@@ -207,3 +207,244 @@ impl<E: MultiMillerLoop> Add for FflonkBundle<E> {
...
+    }
+
+    /// Sub-bundle count as the `u8` written on the wire. The count is
+    /// serialized in a single byte, so a commitment may hold at most 255
+    /// sub-bundles.
+    fn sub_bundle_count_u8(&self) -> u8 {
+        u8::try_from(self.0.len()).expect("FflonkCommitment holds more than 255 sub-bundles.")
+    }
+
+    /// Asserts this commitment is a single sub-bundle (used by `Add`/`Mul`
```
remalue, line 230: "Has" a single sub-bundle?

proofs/src/poly/pcs/commitment.rs
remalue: I am in favor of a simplification of the types here, but haven't put much thought into it what makes sense.

irakoton:
Due to the PR being too large, we close this one and will re-open it in several smaller steps. Comments on this PR will be integrated to the new ones. The _n_ th PR will be prefixed by "Fflonk PR _n_".

proofs/src/transcript/mod.rs
irakoton: Order isn't enforced by the trait, although `Input` is implemented in practice by Vec<u8>.

Note however that the trait itself doesn't require a deterministic order in the iteration, although we need it in practice so that the verifier and the prover agree on said order. I'll add a comment just in case, although in practice it'll likely just be `Vec`.

proofs/src/utils/arithmetic.rs
irakoton: Different lengths are needed. Short story is that multi_open combines polynomials after kate-dividing each q_poly by its point set, and since point sets have different cardinalities the resulting f-polys have different degrees. I'm not sure the reasoning apploes when the feature `fewer-point-sets` is on though, but in general we need that.

proofs/src/poly/pcs/commitment.rs
irakoton: I think the panic is better. Or rather, I can replace it by `unreachable`, this is a case that shouldn't be accessible unless my code is bugged.

proofs/src/poly/pcs/commitment.rs
irakoton: Following a recommendation from Miguel, FflonkBundle and FflonkCommitment will be collapsed in one type, so this issue will disappear.

proofs/src/poly/pcs/commitment.rs
irakoton: Yes, it must be a power of two (mostly to compute roots of unity easily). Added a comment for clarity.

proofs/src/poly/pcs/partition.rs
irakoton: For now yes. When we fully integrate fflonk, I expect we remove them though.

proofs/src/poly/pcs/partition.rs
irakoton: now solved due to removal of BundleFamily

proofs/src/poly/pcs/partition.rs
irakoton: Fflonk is relatively independent of quotient split (in the sense that you can have quotient split or not), although it's better not to split since we already need the extended SRS anyway.

For linearisation, it's more complex. I think they're a bit antagonistic, typically linearisation prevents some polynomials from being combined (and complexifies the FflonkCommitment type too). Maybe they are not inherently incompatible but I'm not sure.

proofs/src/poly/pcs/partition.rs
irakoton: In practice it is not completely known at compile time, it depends of the circuit. Most of the time it'll indeed be `1 << FFLONK_T_MAX_LOG`, but it can be smaller for large circuits.

The idea is that you can only a circuit / fflonk configuration if `k` (from the circuit) verifies `k` + log2(t_max) <= some bound (2 adicity of the field mostly). Since there is a `k_max`, it would mean you can only handle circuits up to `k_max - FFLONK_T_MAX_LOG`. So in the functions you're quoting, `t_max` is more like an "effective t_max", i.e., `1 << k_effective` with something like `k_effective = max(FFLONK_T_MAX_LOG, k - k_max)`.

proofs/src/utils/arithmetic.rs
irakoton: It's a way to implicitly call `deref`. I needed to be able to slice `poly`, and `Deref` was implemented for the type. I could make an explicit call to `deref()` so that it's less confusing, but the other suggestion you made in the comment below is better IMO.

proofs/src/poly/pcs/commitment.rs
irakoton: I assume we could, but that's not trivial. If we want to add two combined polys, wouldn't we need to be able to retrieve the effective individual polys that have been combined? I have the impression that it's the same research question as making linearization and fflonk compatible.

proofs/src/poly/pcs/commitment.rs
irakoton: For this specific case, it's true. Added for the future PRs.

proofs/src/poly/pcs/commitment.rs
irakoton: it was, but code has been removed per another remark. Good catch 👍

proofs/src/poly/pcs/commitment.rs
irakoton: This indeed rings a bell, but we have it in the other PCS, even in `main` unless I'm tripping.

proofs/src/utils/arithmetic.rs
irakoton: enforced it now to be consistent with the analogue site from earlier.

proofs/src/poly/pcs/commitment.rs
irakoton: I didn't manage to make it compile with [0], seems like it causes some ownership issues. But `into_single` seems to be an indeed more succinct solution.

proofs/src/poly/pcs/commitment.rs
irakoton: same as above

proofs/src/poly/pcs/commitment.rs
irakoton: It was to follow the paper's name, but `fflonk_bundle` is probably cleaner, we have other `g`s in the repo after all, and it's not very explicit.

proofs/src/poly/pcs/commitment.rs
irakoton: there was no `self`, but ok it could make the code more readable.

proofs/src/poly/pcs/partition.rs
irakoton: I realise the phrasing suggests something I didn't mean. The variants happen not to be mixed in the code, but it's not a fundamental limitation. It's just that, in the current code, most of the variants have some other, independent properties that may prevent them from being mixed (different phase or different base typically). I think it's simpler to keep things like that at this point, and sit down at some point in the future to mix whatever can further be mixed.

proofs/src/poly/pcs/partition.rs
irakoton: I removed this type. The underlying limitation stays to some extent for the moment, but is basically a conservative heuristics as you mentioned. The idea is that some labels cannot be bundled at all (linearisation), and that some others also cannot for other reasons (different SRS base, different phase).

The changes to allow for mixing labels is relatively very small and localised (commit calls have to be regrouped + `partition` has to allow it. Just a few lines) but impacts the protocols (through the commit calls). Since not all labels can be mixed, I think we should stick to the non mixing policy for now, and decide properly after integration what we want to mix.

proofs/src/poly/pcs/bundle_expansion.rs
irakoton: I meant "opening on a set of points that is larger than necessary" (typically when you bundle polys that are opened at different points, you have to open the combination at the union of all points). I removed the "over", the sentence is just as clear like that.

proofs/src/poly/pcs/commitment.rs
irakoton: it doesn't. But with the recently opened PR, this becomes a non-issue since the count has been made useless and thus removed.

If one day we need this again, I agree we need to think again at whether the sub-bundle count needs to be hashed. I think it doesn't because it is retrievable from the VK, but I may miss something.

proofs/src/poly/commitment.rs
irakoton: sanity check was the wrong term maybe. I meant the condition that the bunding size is lower than the 2 adicity of the field, and the other related tests that I don't have the full list on the top of my head.

proofs/src/poly/pcs/commitment.rs
irakoton: This is a good point, but for the record, the type we would need is rather something like
```
enum FflonkCommitment<E> {
  Normal(Vec<E::G1, Vec<PolynomialLabel>>),
  Linear(Vec<E::G1>, Vec<PolynomialLabel>),
}
```
(the difference is the type of `Normal`)
You need this to keep the link between which group element corresponds to which set of labels. Otherwise I agree, and this will be part of the fflonk atomic PRs when I open them.

proofs/src/poly/pcs/commitment.rs
irakoton: For future reference, I later opened a PR for that.

proofs/src/poly/pcs/mod.rs
irakoton: refactored it.

proofs/src/poly/pcs/utils.rs
irakoton: it does in a sense, but that's a more general (and expansive) function. Something about computing the whole domain, don't remember exactly the name. The idea was to have a lighter helper here.

proofs/src/poly/pcs/commitment.rs
irakoton: Honestly, reaching the 255 threashold is probably impossible anyway, but yes, previous comments (and open PRs) solve this issue and the count will not be serialised in the concrete integration.

proofs/src/poly/pcs/commitment.rs
irakoton: Needed for Hashable, but true that the code could be just calling the identical one from Serde.

proofs/src/poly/pcs/msm.rs
irakoton: there will be a dedicated PR for restructuring, I agree the folder structure is getting a bit unbalanced.

proofs/src/poly/commitment.rs
irakoton: I'll open a PR for refactoring this. I agree having a clear separation between proofs and circuits is good (applies to several other comments that I'm marking resolved without comment, but they will be handled too).

proofs/src/poly/pcs/commitment.rs
irakoton: I think this one may have been wrong (and the doc comment definitely was). The reason why I wasn't comparing the labels was due to the `label` function actually, in the sense that I was afraid of comparing objects that hadn't been labelled with labelled ones. I'll mark this as resolved since the label infrastructure is now completely changed anyway, but I'll sit down properly for the next version when its integration comes, there are subtleties that I'm possibly still missing.

PR #467, opened 2026-07-07
feat: add fixed-base comb MSM to foreign Edwards chip

kitounliu:
Adds `fixed_base_comb_msm` to the foreign Edwards ECC chip: an interleaved
Lim–Lee fixed-base comb for computing Σ_i k_i · G_i when the bases G_i  are compile-time constants.


circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -1300,6 +1300,102 @@ where
...
+    ///  - Handbook of Applied Cryptography, §14.6.3 (fixed-base comb); Guide to
+    ///    Elliptic Curve Cryptography, §3.3 — treatments in additive/ECC form.
+    #[allow(dead_code)]
+    fn fixed_base_comb_msm<const W: usize>(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        scalars: &[(S::Scalar, usize)],
+        base_vals: &[C::CryptographicGroup],
+    ) -> Result<AssignedForeignEdwardsPoint<F, C, B>, Error> {
+        assert!(W >= 1);
```
davidnevadoc: We could add an upper bound as well here. I think it is safe to say that there is no practical use above 32, that being a very generous upper bound. Probably nothing above 16 is useful.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -1300,6 +1300,102 @@ where
...
+    /// columns. Because the base is constant, the whole table is precomputed
+    /// off-circuit; only the `d - 1` intra-column weights cost in-circuit
+    /// doublings (vs `~num_bits` for the variable-base [`Self::windowed_msm`]).
+    ///
+    /// References:
+    ///  - C. H. Lim and P. J. Lee, "More Flexible Exponentiation with
+    ///    Precomputation", CRYPTO 1994 (LNCS 839) — original comb method.
+    ///  - Handbook of Applied Cryptography, §14.6.3 (fixed-base comb); Guide to
+    ///    Elliptic Curve Cryptography, §3.3 — treatments in additive/ECC form.
+    #[allow(dead_code)]
```
davidnevadoc: I think we have 2 options here. 
Either remove the the `dead_code` lint and make the function public.

Or keep this function as internal and create a new public function `fixed_base_msm` that calls this one with a fixed `W` (the most performant). 

Both solutions look accetable to me, so your call.

davidnevadoc (commented):
Great improvement!
Left some minor comments to add the finishing touches. After that, it is ready to merge. Good job 🚀

circuits/src/ecc/foreign/edwards_chip.rs
kitounliu: added upperbound W <= 16

circuits/src/ecc/foreign/edwards_chip.rs
kitounliu: Thanks! Both options are simple and avoid major changes to msm. I went for option 2 to match with the style of the existing msm.

circuits/src/ecc/foreign/edwards_chip.rs
```diff
@@ -1300,6 +1300,128 @@ where
...
+    ///
+    /// Fixed-base multi-scalar multiplication `sum_b scalars[b] * base_vals[b]`
+    /// for compile-time constant bases. This is the public entry point; it
+    /// delegates to [`Self::fixed_base_comb_msm`] with the comb width `W` fixed
+    /// to the most performant setting for full-width scalar fields.
+    ///
+    /// Each scalar is paired with a bit bound `num_bits`; passing a tight bound
+    /// (rather than the full scalar-field width) reduces the number of comb
+    /// columns and thus the in-circuit cost.
+    pub fn fixed_base_msm(
```
miguel-ambrona, line 1336: We tend to document what functions do, not how they are implemented.
Implementation details can be added through `//` instead of `///`.

davidnevadoc (approved):
Thanks for the improvement! 🚀 
LGTM

PR #482, opened 2026-07-23
v2 fixes

davidnevadoc:
- Add a check to ensure committed instances match the expected length. If they do not, the verifier returns `Error::InvalidInstances`


iquerejeta (commented):
We need to bump the versions of midnight-circuits and midnight-zk-stdlib

iquerejeta (approved):

PR #487, opened 2026-07-30
[Fflonk PR1] Refining protocol and PCS primitives to prepare for fflonk

### LINKED ISSUE #497: Extend PCS trait
### Proposed feature

Add generic functions to the `PolynomialCommitmentScheme` trait to account for operations that will change in the protocol with Fflonk integration.

### User need

Integration of Fflonk that will involve no protocol change from KZG.

### Expected benefit

_No response_

### Details

_No response_

irakoton:
Adds 3 functions to the `PolynomialCommitmentScheme` trait to prepare for fflonk, and call them in the protocol. These functions behave slightly differently for fflonk and kzg, hence the need to account for this at the trait level, so that the protocol can remain identical for the two implementations.

### DIFF
--- a/proofs/src/poly/commitment.rs
+++ b/proofs/src/poly/commitment.rs
@@ -68,6 +68,33 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
         Self::commit_many(params, &[polynomial], &[label])
     }
 
+    /// Squeeze the evaluation point used by the protocol to open committed
+    /// polynomials. The default implementation simply squeezes a challenge,
+    /// but specific PCS may require squeezing over sets of challenges
+    /// verifying certain properties, for example fflonk requires the
+    /// evaluation point to be a `t`-th power in the field.
+    ///
+    /// The protocol must squeeze evaluation points through this method.
+    fn squeeze_evaluation_point<T: Transcript>(transcript: &mut T) -> F
+    where
+        F: Sampleable<T::Hash>,
+    {
+        transcript.squeeze_challenge()
+    }
+
+    /// Multiplicative blow-up factor by which `params.g_monomial_size()` must
+    /// exceed `2^k` (the circuit's Lagrange-domain size) for this PCS to
+    /// commit every polynomial it produces at the requested circuit size.
+    /// Returns `1` when no extension is needed.
+    ///
+    /// `cs_degree` is the constraint system's `cs.degree()`. Schemes that
+    /// commit to a single combined polynomial (e.g. `single-h-commitment`,
+    /// fflonk's bundles) factor that into their requested blow-up.
+    fn srs_monomial_blowup(cs_degree: usize) -> usize {
+        let _ = cs_degree; // Just to avoid a clippy warning.
+        1
+    }
+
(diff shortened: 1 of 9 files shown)


proofs/src/poly/commitment.rs
```diff
@@ -68,6 +68,33 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
         Self::commit_many(params, &[polynomial], &[label])
     }
 
+    /// Squeeze the evaluation point used by the protocol to open committed
+    /// polynomials. The default implementation simply squeezes a challenge,
+    /// but specific PCS may require squeezing over sets of challenges
+    /// verifying certain properties, for example fflonk requires the
+    /// evaluation point to be a `t`-th power in the field.
```
remalue: "verifiying" - > "satisfying"?

Also, not clear what "squeezing over sets of challenges means". In FFLONK it's still a single eval challenge that is squeezed (but with certain properties, as you stated). The eval points are derived from it.

zk_stdlib/src/utils/plonk_api.rs
```diff
@@ -261,28 +261,26 @@ pub enum SrsSource {
 /// Loads an SRS (over BLS12-381) for the given circuit size `k` and
 /// constraint-system degree `cs_degree`.
 ///
-/// Without the `single-h-commitment` feature, the monomial and Lagrange bases
-/// have the same size `2^k`. With the feature enabled the monomial basis is
-/// extended to `k + ceil(log2(cs_degree - 1))` so that it can hold the full
-/// quotient polynomial, while the Lagrange basis is kept at size `2^k`.
+/// The monomial basis is sized via the active PCS's
+/// [`PolynomialCommitmentScheme::srs_monomial_blowup`]: a blow-up factor of
+/// `1` keeps the monomial basis at the Lagrange size `2^k`; a blow-up of `B`
```
remalue, line 266: "blow-up factor of `1`" -> "blow-up factor of 0 or 1"?

proofs/src/poly/kzg/mod.rs
```diff
@@ -609,7 +622,7 @@ mod tests {
         };
 
         let result =
-            KZGCommitmentScheme::multi_prepare(&queries.collect::<Vec<_>>(), &mut transcript)
+            KZGCommitmentScheme::multi_prepare(&queries.collect::<Vec<_>>(), 4, &mut transcript)
```
remalue: Pass it as a parameter to the `verify` function instead of hard-coding here?

remalue (dismissed):
LGTM

iquerejeta (dismissed):

proofs/src/poly/commitment.rs
irakoton: Good point, I rephrased accordingly.

zk_stdlib/src/utils/plonk_api.rs
irakoton: True. I think it's however a bit cleaner to assert `!= 0` and then `== 1` instead of `<= 1`.

iquerejeta (approved):

PR #491, opened 2026-08-04
[Fflonk PR2] Removing Labeleable trait

### LINKED ISSUE #498: Removing `Labelable` trait
### Summary

Removes the `Labelable` trait from the commitment interface to clean the general interface and simplify Fflonk's integration.

_No response_

### Context / notes

_No response_

irakoton:
Removes the Labelable trait and replaces it by reading functions that directly take the slice of the labels of the polynomials that are intended to be read. This will remove the necessity to serialise the bundle sizes when implementing fflonk, while also making the general interface of the transcript reading cleaner.

The diff is relatively atomic. First commit removes the incriminated trait and adds replacement functions to `PolynomialCommitmentScheme`. Subsequent commits update the doc and the call sites.


iquerejeta (commented):
Looks good to me. I'd like @miguel-ambrona to have a look at it, as he included the `Labelable` trait, to make sure we are not missing anything.

proofs/src/poly/kzg/mod.rs
```diff
@@ -119,6 +119,41 @@ where
...
+        Self::Commitment: Hashable<T::Hash>,
+    {
+        // KZG commits each polynomial independently, so a commitment to
+        // `labels.len()` polynomials is `labels.len()` points read (and hashed)
+        // one after another, each tagged with its label.
+        let inners = labels
+            .iter()
+            .map(|label| {
+                let com: KZGMultiCommitment<E> = transcript.read()?;
+                Ok(com.into_single().label(label.clone()))
```
miguel-ambrona: It would be good to also get rid of this `.label` method. Is it possible? You can call a constructor here that also takes the label.

proofs/src/poly/kzg/mod.rs
```diff
@@ -119,6 +119,41 @@ where
...
+            })
+            .collect::<io::Result<Vec<_>>>()?;
+        Ok(KZGMultiCommitment(inners))
+    }
+
+    fn deserialize_commitment<R: Read>(
```
miguel-ambrona, line 149: Are both "read_commitment" and "deserializable_commitment" necessary? Can we get rid of one of them?

proofs/src/poly/commitment.rs
```diff
@@ -132,24 +149,6 @@ pub trait PolynomialCommitmentScheme<F: PrimeField>: Clone + Debug {
...
-    /// Attaches the given labels to this commitment, one per polynomial,
-    /// replacing any existing labels (including [`PolynomialLabel::NoLabel`]).
-    ///
-    /// # Panics
-    ///
-    /// Panics if `labels.len() != self.length()`.
-    fn label(self, labels: &[PolynomialLabel]) -> Self;
-
-    /// Returns the number of polynomials held by this commitment.
-    fn length(&self) -> usize;
```
miguel-ambrona, line 150: I wonder if `length` should still be accessible. Maybe in a different trait, maybe directly from the PCS trait.

proofs/src/plonk/permutation/verifier.rs
```diff
@@ -42,12 +39,11 @@ impl Argument {
         let permutation_product_commitments = self
             .columns
             .chunks(chunk_len)
-            .map(|_| transcript.read::<CS::Commitment>())
-            .collect::<Result<Vec<_>, _>>()?
-            .into_iter()
             .enumerate()
-            .map(|(i, c)| c.label(&[PolynomialLabel::PermutationAccumulator(i)]))
-            .collect();
+            .map(|(i, _)| {
```
miguel-ambrona: Since you do not use the second argument, it may be better to iterate over `0..self.columns.len().div_ceil(chunk_len)` instead of over `self.columns.chunks(chunk_len)`.

proofs/src/poly/kzg/commitment.rs
```diff
@@ -283,37 +283,21 @@ where
...
+/// [`PolynomialLabel::NoLabel`] and are labeled by the reader.
 impl<E: MultiMillerLoop> ProcessedSerdeObject for KZGMultiCommitment<E>
 where
     E::G1: Default + ProcessedSerdeObject,
 {
     fn read<R: io::Read>(reader: &mut R, format: SerdeFormat) -> io::Result<Self> {
-        Ok(Self(vec![
-            <KZGCommitment<E> as ProcessedSerdeObject>::read(reader, format)?,
-        ]))
+        Ok(Self(vec![KZGCommitment::<E>::read(reader, format)?]))
```
miguel-ambrona: Hum. Do we need to implement this trait? It seems we do not wanna use this `read` function. We wanna pass the labels instead.

circuits/src/verifier/permutation.rs
```diff
@@ -65,12 +65,15 @@ pub(crate) fn read_product_commitments<S: SelfEmulation, PCS: InCircuitPCS<S>>(
         .permutation()
         .get_columns()
         .chunks(chunk_len)
-        .map(|_| PCS::read_commitment(transcript_gadget, layouter, 1))
-        .collect::<Result<Vec<_>, _>>()?
-        .into_iter()
         .enumerate()
-        .map(|(i, c)| c.label(&[PolynomialLabel::PermutationAccumulator(i)]))
-        .collect();
+        .map(|(i, _)| {
```
miguel-ambrona: Same here, since the second argument is unused, we can polish the initial iterator.

miguel-ambrona (dismissed):
LGTM.

Only minor comments. The most demanding ones do not need to be addressed now (they can be converted into issues maybe). Let's get Flonk working first!

circuits/src/verifier/pcs.rs
```diff
@@ -101,13 +99,13 @@ pub trait InCircuitPCS<S: SelfEmulation>: Sized + Clone + Debug {
         label: PolynomialLabel,
     ) -> Result<Self::AssignedCommitment, Error>;
 
-    /// Reads one commitment (to `length` polynomials) from the proof
-    /// transcript. Commitments are not length-prefixed on the wire, so the
-    /// caller supplies the polynomial count (`1` unless batched).
+    /// Reads one commitment (to `labels.len()` polynomials) from the proof
+    /// transcript, tagging each polynomial with its label. Commitments are not
+    /// length-prefixed on the wire; the label count gives the polynomial count.
```
remalue: Nit: Maybe the first sentence

"Reads one commitment to `labels.len()` polynomials from the proof transcript, tagging each polynomial with its label."

suffices? The second part of the doc basically repeats the first part. Or is it obligatory to know that commitments are not length-prefixed?

proofs/src/poly/kzg/mod.rs
irakoton: Yes. Replaced the label function by a constructor call.

proofs/src/poly/kzg/mod.rs
irakoton: I'm not sure. I wanted to remove one at some point too, but didn't really managed to do it. This is not really specific to fflonk, basically the two correspond to the two flavours of `read` we have in the library (`Hashable` VS `ProcessedSerdeObject`). Maybe we can solve this in the future, but I expect it's a more substantial refactor of the library. I'll open an issue.

proofs/src/poly/commitment.rs
irakoton: I'm not sure, but I'm leaning on no. Since we don't serialise the length in question, having access to the length in the code is a bit weird IMO. I'm wondering if it's not even dangerous to some extent (in that it allows you to do things with an object that is in general not in the transcript itself). I think it's fine to have an associated function for the specific implementation of KZG and/or Fflonk, but not something generic at the trait level.

proofs/src/poly/kzg/commitment.rs
irakoton: We need the trait at least for the `write`, but I agree that you don't want the `read` to be called. I replaced its code by `unimplemented!` with an error message indicating that you should use `deserialize_commitment` instead, which is the label-passing equivalent. Maybe in the future we can find a cleaner way to require the write` function, but I think it's fine for now.

circuits/src/verifier/pcs.rs
irakoton: I agree it's better. I made a pass on the repo to remove mentions of the (absence of) length prefix.

remalue (dismissed):
LGTM!

The only point that caugt my attention is the one that @miguel-ambrona mentioned, about `read_commitment` (reads from proof and hashes to transcript) vs `deserialize_commitment` (reads from keys, but doesn't hash).

While I don't have a proposal for how to best streamline (there will be an issue IIUC), for now making the difference more explicit in the doc could help here. As you see fit.

irakoton:
For reference, the issue regarding the interface for reading/deserialising commitments is there https://github.com/midnightntwrk/midnight-zk/issues/495.

remalue (approved):

PR #493, opened 2026-08-12
[Fflonk PR3] Batching `commit` calls

### LINKED ISSUE #499: Cross-argument commitment batching
### Summary

Rewire how commitments are called in the protocol so that each phase is performing a single call to `commit_many` instead of many individual calls across the phase. Needed for Fflonk.

_No response_

### Context / notes

_No response_

irakoton:
Prepares for fflonk implementation by changing how the calls to `commit`/`commit_many` are done. Right now they're called individually whenever a commit has to be made. Now a single `commit_many` call is performed at the end of the corresponding phase.

The PR commit history is basically 1 commit for each protocol phase (no pun intended).

NB: it was a bit confusing why, in commit 4, I reordered some commitments as said in the commit message, but that you cannot observe a change of goldenfiles. I think the reason is that for the proof bytes to change, we would need examples with ≥2 logup arguments (none for now it seems).

### DIFF
--- a/circuits/src/verifier/lookup.rs
+++ b/circuits/src/verifier/lookup.rs
@@ -75,36 +75,76 @@ pub(crate) fn read_multiplicities<S: SelfEmulation, PCS: InCircuitPCS<S>>(
         .collect())
 }
 
-impl<S: SelfEmulation, PCS: InCircuitPCS<S>> CommittedMultiplicities<S, PCS> {
-    pub(crate) fn read_commitment(
-        self,
-        argument_index: usize,
-        nb_flattened: usize,
-        layouter: &mut impl Layouter<S::F>,
-        transcript_gadget: &mut TranscriptGadget<S>,
-    ) -> Result<Committed<S, PCS>, Error> {
-        let helper_polys = (0..nb_flattened)
-            .map(|j| {
-                PCS::read_commitment(
-                    transcript_gadget,
-                    layouter,
-                    &[PolynomialLabel::LogupHelper(argument_index, j)],
-                )
-            })
-            .collect::<Result<Vec<_>, Error>>()?;
-        let accumulator = PCS::read_commitment(
-            transcript_gadget,
-            layouter,
-            &[PolynomialLabel::LogupAggregator(argument_index)],
-        )?;
+/// Partial state between the batched helper read and the batched aggregator
+/// read.
+#[derive(Clone, Debug)]
+pub(crate) struct Helpers<S: SelfEmulation, PCS: InCircuitPCS<S>> {
+    argument_index: usize,
+    multiplicities: PCS::AssignedCommitment,
+    helper_polys: Vec<PCS::AssignedCommitment>,
+}
 
-        Ok(Committed {
-            argument_index,
-            multiplicities: self.multiplicities,
-            helper_polys,
-            accumulator,
-        })
+/// Reads the batched helper commitments for all logup arguments.
+pub(crate) fn read_helpers<S: SelfEmulation, PCS: InCircuitPCS<S>>(
+    args_with_multiplicities: Vec<(usize, usize, CommittedMultiplicities<S, PCS>)>,
+    layouter: &mut impl Layouter<S::F>,
+    transcript_gadget: &mut TranscriptGadget<S>,
+) -> Result<Vec<Helpers<S, PCS>>, Error> {
+    // Build the full label set in (arg, chunk) order, matching the prover's
+    // flat iteration.
+    let mut labels: Vec<PolynomialLabel> = Vec::new();
+    for (arg, nb_chunks, _) in &args_with_multiplicities {
+        for chunk_idx in 0..*nb_chunks {
+            labels.push(PolynomialLabel::LogupHelper(*arg, chunk_idx));
+        }
     }
+    if labels.is_empty() {
+        // No helpers anywhere — the prover skipped the write.
+        return Ok(args_with_multiplicities
+            .into_iter()
+            .map(|(argument_index, _, m)| Helpers {
+                argument_index,
+                multiplicities: m.multiplicities,
+                helper_polys: Vec::new(),
+            })
+            .collect());
+    }
+    let shared = PCS::read_commitment(transcript_gadget, layouter, &labels)?;
+    Ok(args_with_multiplicities
+        .into_iter()
+        .map(|(argument_index, nb_chunks, m)| Helpers {
+            argument_index,
+            multiplicities: m.multiplicities,
+            helper_polys: vec![shared.clone(); nb_chunks],
+        })
+        .collect())
+}
+
+/// Reads the batched aggregator commitment for all logup arguments and
+/// assembles one [`Committed`] per arg. Mirrors the off-circuit
+/// `logup::verifier::read_aggregators`.
+pub(crate) fn read_aggregators<S: SelfEmulation, PCS: InCircuitPCS<S>>(
+    helpers: Vec<Helpers<S, PCS>>,
+    layouter: &mut impl Layouter<S::F>,
+    transcript_gadget: &mut TranscriptGadget<S>,
+) -> Result<Vec<Committed<S, PCS>>, Error> {
+    if helpers.is_empty() {
+        return Ok(Vec::new());
+    }
+    let labels: Vec<_> = helpers
+        .iter()
+        .map(|h| PolynomialLabel::LogupAggregator(h.argument_index))
+        .collect();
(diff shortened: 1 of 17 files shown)


proofs/src/poly/kzg/mod.rs
```diff
@@ -161,6 +161,20 @@ where
...
+    where
+        Self::Commitment: Hashable<T::Hash>,
+    {
+        // KZG commits each polynomial independently.
+        for inner in &commitment.0 {
+            transcript.write(&KZGMultiCommitment(vec![inner.clone()]))?;
+        }
+        Ok(())
+    }
+
```
iquerejeta, line 176: You should modify the implementation of `Hashable` instead (and not add this function to the PCS trait).

proofs/src/plonk/logup/prover.rs
```diff
@@ -392,6 +392,66 @@ impl<F: WithSmallOrderMulGroup<3>> Evaluated<F> {
     }
 }
 
+/// Computes the multiplicities of every logup argument and writes one batched
+/// commitment to all of them. Mirrors the verifier's `read_multiplicities`.
```
iquerejeta, line 362: I would instead have a function that computes the multiplicities, and then you batch_commit outside of the call. This will facilitate a future transition where we batch commit arguments from the same phase.

proofs/src/plonk/logup/verifier.rs
```diff
@@ -49,56 +49,33 @@ pub struct Evaluated<F: PrimeField, CS: PolynomialCommitmentScheme<F>> {
...
-    where
-        CS::Commitment: Hashable<T::Hash>,
-    {
-        let multiplicities = CS::read_commitment(
-            transcript,
-            &[PolynomialLabel::LogupMultiplicities(argument_index)],
-        )?;
-        Ok(CommittedMultiplicities { multiplicities })
+/// Reads the batched multiplicities commitment for all logup arguments in one
+/// transcript entry (one point per argument) and hands each argument a clone of
```
iquerejeta: Rather than 'one point per argument', shouldn't it be 'one commitment per argument'?

proofs/src/plonk/verifier.rs
```diff
@@ -88,12 +85,22 @@ where
...
+        .map(|(i, (batch, m))| {
+            (
+                ChunkedArgRef {
+                    argument_index: i,
+                    num_chunks: batch.num_chunks(),
+                },
+                m,
+            )
+        })
+        .collect();
```
iquerejeta, line 111: here you need to read the helper and aggregator commitments of the logup argument. When is that happening?

circuits/src/verifier/lookup.rs
```diff
@@ -57,18 +57,22 @@ pub(crate) struct Evaluated<S: SelfEmulation, PCS: InCircuitPCS<S>> {
...
+    if num_args == 0 {
+        return Ok(Vec::new());
+    }
+    let labels: Vec<_> = (0..num_args).map(PolynomialLabel::LogupMultiplicities).collect();
+    let shared = PCS::read_commitment(transcript_gadget, layouter, &labels)?;
+    Ok((0..num_args)
+        .map(|_| CommittedMultiplicities {
+            multiplicities: shared.clone(),
+        })
+        .collect())
```
iquerejeta, line 75: Here we should return a single commitment.

circuits/src/verifier/verifier_gadget.rs
```diff
@@ -357,28 +357,25 @@ impl<S: SelfEmulation> VerifierGadget<S> {
...
         let gamma = transcript.squeeze_challenge(layouter)?;
 
         let permutation_committed =
-            // Hash each permutation product commitment
             permutation::read_product_commitments(layouter, &mut transcript, cs)?;
 
-        let lookups_committed = multiplicities_committed
+        // Read one batched helper commitment for all (arg, chunk) helpers, then
+        // one batched aggregator commitment.
+        let args_with_multiplicities: Vec<_> = multiplicities_committed
```
iquerejeta, line 377: and instead of iterating over this (this will now be a single element), we just repeat the element `n` times, where `n` is the number of arguments.

proofs/src/plonk/logup/prover.rs
```diff
@@ -452,6 +419,76 @@ where
...
+            helper_polys.push(domain.lagrange_from_vec(h.clone()));
+            labels.push(PolynomialLabel::LogupHelper(c.argument_index, chunk_idx));
+        }
+    }
+
+    if helper_polys.is_empty() {
+        return Ok(());
+    }
+
+    let helper_refs: Vec<_> = helper_polys.iter().collect();
```
iquerejeta: seems annoying that we need to do this every time we call commit_many - maybe a misdesign of `commit_many`s signature?

proofs/src/plonk/prover.rs
```diff
@@ -164,55 +171,33 @@ where
...
-        for h_commitment in helper_commitments {
-            transcript.write(h_commitment)?;
-        }
-        transcript.write(&c.aggregator_commitment)?;
-    }
-    let lookups: Vec<logup::prover::Committed<F>> = computed
+    let computed = logup_computed?;
+
+    commit_helpers(params, pk, &computed, transcript)?;
+    commit_aggregators::<F, CS, T>(params, &computed, transcript)?;
```
iquerejeta, line 190: IMO this looks better - have this calls explicitly in the plonk protocol. That's what I meant in my previous comment on the multiplicities.

proofs/src/plonk/logup/verifier.rs
```diff
@@ -76,8 +91,82 @@ where
...
+            })
+            .collect());
+    }
+    let shared = CS::read_commitment(transcript, &labels)?;
+    Ok(args_with_multiplicities
+        .into_iter()
+        .map(|(arg, m)| HelpersOnly {
+            argument_index: arg.argument_index,
+            multiplicities: m.multiplicities,
+            helper_polys: vec![shared.clone(); arg.num_chunks],
```
iquerejeta, line 136: given that we are committing to all helper polys in a shared commitment, why keep the `Vec<>` structure?

iquerejeta (commented):
Maybe some comments are confusions from reviewing commit by commit. 

However, I would like to propose designing arguments such that the multi commit is called in the plonk protocol (rather than in the corresponding argument calls). This will facilitate moving towards a shared commitment over several arguments in the future.

proofs/src/poly/kzg/mod.rs
```diff
@@ -161,6 +161,20 @@ where
...
+    fn write_commitment<T: Transcript>(
+        transcript: &mut T,
+        commitment: &Self::Commitment,
+    ) -> io::Result<()>
+    where
+        Self::Commitment: Hashable<T::Hash>,
+    {
+        // KZG commits each polynomial independently.
+        for inner in &commitment.0 {
+            transcript.write(&KZGMultiCommitment(vec![inner.clone()]))?;
```
miguel-ambrona, line 172: Why creating a `KZGMultiCommitment`? Can't we just write the point directly?

circuits/src/verifier/lookup.rs
```diff
@@ -57,18 +57,22 @@ pub(crate) struct Evaluated<S: SelfEmulation, PCS: InCircuitPCS<S>> {
...
-    let multiplicities = PCS::read_commitment(
-        transcript_gadget,
-        layouter,
-        &[PolynomialLabel::LogupMultiplicities(argument_index)],
-    )?;
-    Ok(CommittedMultiplicities { multiplicities })
+) -> Result<Vec<CommittedMultiplicities<S, PCS>>, Error> {
+    if num_args == 0 {
+        return Ok(Vec::new());
+    }
```
miguel-ambrona, line 68: Hum, I would return an error if 0 is passed.

circuits/src/verifier/lookup.rs
miguel-ambrona, line 75: This is a bit weird, but I understand it can be a first step.

I mean, the fact that all arguments have the same shared commitment.
Maybe it makes sense that `CommittedMultiplicities` also contains a field for the label (which will be different across argument).

miguel-ambrona: In general, it feels that the right approach is that arguments do not return commitments anymore. Instead they return labels and polynomials and those polynomials are committed to together in the main protocol at each phase.

miguel-ambrona (commented):
I'm not sure about this approach.

It feels better to do what I suggested: remove the "Committed" struct from all arguments. Instead, arguments would return a struct that contains a polynomial (and its label). The commit operation would take place in the main protocol at each phase.

PR #501, opened 2026-08-13
[Fflonk PR4] [REFACTOR] Clean PCS file tree

irakoton:
With the growing number of PCS-related files, the structure has become a bit outdated and chaotic. This PR mostly moves files around, but also removes the circuit-related argument in `multi_prepare` (integrating feedback from the reference branch).

The diff is big, but 99% of it is just moving files and updating the imports accordingly.


iquerejeta (dismissed):
Looks good - just a nit. It is not always necessary to update the changelogs (even if the CI fails). For example, the aggregation crate only changed the imports. No need to add a changelo entry for that. No need to change it here, just for future reference

remalue:
In the numbering of FFLONK PRs, is this  `[Fflonk PR4]`?

PR #506, opened 2026-08-14
chore: standard issue templates + label-automation dispatchers

### LINKED ISSUE #505: Standardize issue templates and label automation
Adopt the org-standard YAML issue forms (native `type:`, `status:untriaged`, this repo's `component:*` label, component dropdown where the repo spans several components) and the label-automation dispatchers calling the org reusables (untriaged stamp, Origin default, untriaged-drop on assign, bot:ai-assisted PR backstop, component-from-form). Part of the org label standardization; automation requires midnightntwrk/midnight-iac#314's apply (Actions access level on .github).

chrisferry:
Org-standard YAML issue forms (native `type:` Bug/Feature/Task, `status:untriaged`, this repo's `component:*` default, dropdown where multi-component) replacing the legacy `.md` templates, plus SHA-pinned dispatchers for the org reusable label workflows.

**Merge gate:** dispatchers resolve only after midnightntwrk/midnight-iac#314's apply (Actions access level on the private `.github`) — verified via a re-run of midnight-zk's previously failing dispatcher before this rolls.

Closes #505


chrisferry:
Scope added per review request: cleared the open code-scanning findings that live in workflow files. 22 alerts across three defect classes, all pre-existing — none were in the dispatchers this PR adds.

**Token-Permissions (high × 9)** — `ci.yaml`, `circuits_examples.yml`, `aggregation_examples.yml`, `publish-crate.yml`, and the five `changelog_*.yml` had no top-level `permissions:`, so every job ran with the default broad token. Each now declares least privilege. I checked every job in those files for anything needing a write scope (gh CLI, github-script, comment/push/release actions) — there are none, so `contents: read` is sufficient; the changelog workflows also get `pull-requests: read` for the enforcer.

`publish-crate.yml` needed care: its job-level block **replaces** the top-level one rather than merging, so declaring only `id-token: write` there left `contents` unset for `checkout`. It now declares both.

**Mutable/unpinned action tags (× 4)** — `dtolnay/rust-toolchain@nightly`, `actions/cache@v6` (×2), `rust-lang/crates-io-auth-action@v1.0.5`, `Swatinem/rust-cache@v2`, all pinned to full 40-char SHAs with a version comment. `actions/cache` is pinned to the same SHA the rest of the repo already uses. Note `dtolnay/rust-toolchain` derives the toolchain from the ref when referenced by branch, so pinning to a SHA requires passing the (required) `toolchain: nightly` input explicitly — done, otherwise the step would have started failing.

**Shell injection (error × 1)** — `.github/actions/cargo-test/action.yaml` interpolated `${{ inputs.package }}` / `${{ inputs.args }}` directly into `run:`, where they become command text. Now bound via `env:` and referenced as shell variables. `$ARGS` stays deliberately unquoted since it carries multiple cargo flags, with a scoped shellcheck disable and a comment saying why.

Two findings I did **not** change, flagged instead:

- `ci.yaml:17` references `steps.cache-srs.outputs.cache-hit`, but that step is `id: restore-srs`. `actionlint` confirms the property is undefined, so `srs-cache-hit` is always empty and any gate on it is dead. Pre-existing logic bug, unrelated to security — worth its own issue.
- `ci.yaml` uses `ubuntu-latest` and `ubuntu-latest-16-core-x64`. Our convention is pinned runner images rather than `*-latest`; that is a broader change than this PR should carry.
- Alert #40 points at `.github/workflows/checkmarx.yaml`, which does not exist on this branch — a stale alert that needs dismissing rather than a code fix.

iquerejeta (dismissed):

night-warden (dismissed):

erik-axeljakobsson (dismissed):

erik-axeljakobsson (approved):

iquerejeta (approved):

PR #508, opened 2026-08-14
Update the crate to Rust edition 2024

irakoton:
Largely automated PR to bump the whole codebase to the latest syntax conventions of Rust (edition 2024). Also bump the required Rust version (was outdated, some parts of the codebase was already using syntax features from 1.88, so I required this version in the whole repo).

The very large majority of the diff are obtained by automated commands (`cargo fix` or `cargo fmt`) + a post-processing to have more idiomatic code (`cargo fix` sometimes generates awkward code). Note that edition 2024 changes the convention on how to format and sort imports, which is why the number of edited files is so big.


Cargo.toml
```diff
@@ -3,7 +3,10 @@ resolver = "2"
 
 members = ["proofs", "curves", "circuits", "aggregation", "zk_stdlib"]
 
-package = { license-file = "LICENSE" }
+[workspace.package]
+license-file = "LICENSE"
+edition = "2024"
+rust-version = "1.88.0"
```
iquerejeta: 1.88.0? in `rust-toolchain.toml` we were already in 1.90.0. Maybe we can update that there as well to a newer version. And given that you are at it, could you change the `components` line of that file to `components = ["rustfmt", "clippy", "rust-analyzer"]`?

remalue: +1 for the adding rust-analyzer to components

irakoton: It's possible, I used 1.88 because it was the minimal config needed to compile the code (bumped from something like 1.76, can't remember exactly). I can upgrade to 1.90 so that everything is alined.

For the record too, I don't think this constrain the CI, and I don't know which version is used in the CI tests. I assume @iquerejeta knows where to check that?

iquerejeta: The CI uses the version in [rust-toolchain.toml](https://github.com/midnightntwrk/midnight-zk/blob/main/rust-toolchain.toml)

circuits/src/ecc/native/edwards_chip.rs
```diff
@@ -638,7 +638,7 @@ impl<C: EdwardsCurve> EccChip<C> {
     }
 
     /// The native gadget carried by this chip.
-    pub fn native_gadget(&self) -> &impl NativeInstructions<C::Base> {
+    pub fn native_gadget(&self) -> &(impl NativeInstructions<C::Base> + use<C>) {
```
iquerejeta: What's this new notation? looks ugly

proofs/src/plonk/logup/verifier.rs
```diff
@@ -141,7 +141,7 @@ impl<F: WithSmallOrderMulGroup<3>, CS: PolynomialCommitmentScheme<F>> Evaluated<
         &self,
         vk: &VerifyingKey<F, CS>,
         x: F,
-    ) -> impl Iterator<Item = VerifierQuery<'_, F, CS>> + Clone {
+    ) -> impl Iterator<Item = VerifierQuery<'_, F, CS>> + Clone + use<'_, F, CS> {
```
iquerejeta: There's a few of these new additions. Are they unavoidable?

iquerejeta (dismissed):
Minor comments - do you know about the new style I commented on? Doesn't look great.

proofs/src/plonk/logup/verifier.rs
irakoton: same as above

circuits/src/ecc/native/edwards_chip.rs
irakoton: Indeed ugly, was introduced by `cargo fix`. I investigated the exact reason why it's introduced, and basically, edition 2024, whenever you introduce a lifetime (even elided, so it includes the `&self` here), it's implicitly bound in the result too. So when tasked to bump from edition 2021 to 2024, `cargo fix` adds a `use` to force the function to use exactly the same lifetimes as with the previous edition.

It appears that, in this specific case, it works without it. In the sense: it compiles even when the `use` is removed (same for the other occurrences). The behaviour of the function _is_ different now though, so there might be degenerate uses of the above function that would return an error now. But as far as I understand, they are very unlikely, and the Rust compiler will suggest the `use<>` as a patch when it happens anyway.

TLDR: I removed it, it works, and if one day it doesn't, the compiler will give us the solution.

irakoton:
Made a pass on the `use<>` (happen to be useless for the current usage of the incriminated functions). I also made a pass on the whole diff with Claude to flag generated code that is bloated, and a couple of sites were cleaned this way. This was limited/targeted to code generated by the `cargo fix`.

iquerejeta (approved):

PR #512, opened 2026-08-20
add cargo lock file to repo

perturbing:
As per [this](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html) and [this,](https://blog.rust-lang.org/2023/08/29/committing-lockfiles/) it is no longer required for libraries to not have a lock file.

I would even think it is wise, as it guards against supply chain attacks.

Also not that [this](https://www.reddit.com/r/rust/comments/1vtdfzb/arrayref_v0310_compromised/) example that just was cought did affect this repo since
```
proc-macro1 v1.0.107
└── arrayref v0.3.10            ← forged: real arrayref 0.3.10 has NO deps
    └── blake2b_simd v1.0.4
        ├── midnight-proofs v0.8.0   (Fiat–Shamir transcript hash)
        ├── midnight-circuits v7.1.0
        ├── midnight-zk-stdlib v2.2.0
        ├── blake2b_halo2 (git rev 257ae18)
        └── sha3-circuit  (git rev 1a3ef7d)
```

for more context see https://www.stepsecurity.io/blog/arrayref-rust-crate-supply-chain-attack#am-i-affected-what-to-do-now


iquerejeta (approved):
Approving - checked that it matches my local build.

PR #513, opened 2026-08-20
Introducing generic Arguments

miguel-ambrona:
Introduce a generic `Argument` to be able to commit to polynomials from different arguments at the same time.
Implement it for the trash argument only so far.


proofs/src/plonk/prover.rs
```diff
@@ -333,13 +332,12 @@ where
...
         instance_polys,
         lookups,
-        trashcans,
+        phase2_committed,
         permutations,
         beta,
         gamma,
         theta,
         trash_challenge,
-        y,
```
iquerejeta, line 342:
```suggestion
        y,
```

proofs/src/plonk/prover.rs
```diff
@@ -399,8 +394,14 @@ where
     );
 
     // Compute linearization polynomial
-    let (lin_poly_non_constant_part, lin_poly_constant_term) =
-        compute_linearization_poly(expressions, pk, y, xn, splitting_factor, quotient_limbs);
+    let (lin_poly_non_constant_part, lin_poly_constant_term) = compute_linearization_poly(
+        expressions,
+        pk,
+        trace.y,
```
iquerejeta:
```suggestion
        y,
```

proofs/src/plonk/trash.rs
```diff
@@ -49,20 +49,18 @@ impl<F: Field> Argument<F> {
...
+        &self,
+        evals_map: &BTreeMap<PolynomialLabel, Vec<argument::Evaluation<F>>>,
         trash_challenge: F,
         advice_evals: &[F],
         fixed_evals: &[F],
         instance_evals: &[F],
-    ) -> impl Iterator<Item = F> + 'a {
+    ) -> impl Iterator<Item = F> {
+        let trash_label = PolynomialLabel::Trash(self.argument_index);
+        let trash_eval = evals_map.get(&trash_label).unwrap()[0].eval();
```
iquerejeta, line 62: I'm concerned of this `unwrap`, as it is the verifier who runs it. However, I think that it can indeed never panic, because the label is defined by the verifier, and in any case it would have failed in the 'reading' phase. Can you confirm that?

iquerejeta (dismissed):
LGTM - looking forward seeing this in the other arguments.

proofs/src/plonk/trash.rs
miguel-ambrona: Yes I think it can never panic. Nevertheless, let's add a comment or GH issue to address it.

miguel-ambrona: https://github.com/midnightntwrk/midnight-zk/issues/516

iquerejeta (approved):

PR #515, opened 2026-08-21
Logup as a generic argument

miguel-ambrona:


proofs/src/plonk/argument/verifier.rs
```diff
@@ -18,16 +20,49 @@ pub struct Committed<F: PrimeField, CS: PolynomialCommitmentScheme<F>> {
...
+        // The prover commits to the group in the labels' `Ord` order (its
+        // polynomials live in a `BTreeMap`), so read them in that order,
+        // whatever order the caller listed them in.
+        let polynomial_labels = BTreeSet::from_iter(labels.iter().cloned());
+        let ordered_labels: Vec<_> = polynomial_labels.iter().cloned().collect();
         Ok(Committed {
-            commitment: CS::read_commitment(transcript, labels)?,
-            polynomial_labels: BTreeSet::from_iter(labels.iter().cloned()),
+            commitment: CS::read_commitment(transcript, &ordered_labels)?,
+            polynomial_labels,
```
iquerejeta: I'm thinking that it may make sense to make this ordering directly inside the read/write function.

iquerejeta (dismissed):
Looks great - thinking how to minimise errors with the ordering of the labels. It can be fixed in the next PR.

proofs/src/poly/kzg/mod.rs
davidnevadoc: This may trigger a panic in the verifiier. 

The main reason is that this `read` is not made through `read_commitment` like the others.
In this case we are calling `read` directly with `<KZGMultiCommitment<E>>` ( not `KZGCommitment`) so we will get a vector which is not guaranteed to be of length 1. It will be for honest proofs, but otherwise, `into_single` panics rather than returning an error.

davidnevadoc:
```
        let f_com = <KZGCommitmentScheme<_> as PolynomialCommitmentScheme<_>>::read_commitment(
            transcript,
            &[PolynomialLabel::Custom("kzg_batch".to_string())],
        )
        .map_err(|_| Error::SamplingError)?;
```

proofs/src/plonk/argument/prover.rs
```diff
@@ -21,6 +22,20 @@ pub(crate) struct Committed<F: PrimeField, B: PolynomialRepresentation> {
...
+    pub fn into_coeff(self, domain: &EvaluationDomain<F>) -> Committed<F, Coeff> {
+        Committed {
+            polys_map: BTreeMap::from_iter(
+                self.polys_map
+                    .into_par_iter()
+                    .map(|(label, p)| (label, B::self_to_coeff(domain, p)))
+                    .collect::<Vec<_>>()
+                    .into_iter(),
+            ),
+        }
```
irakoton: Pretty the following code work, even with parallel iterators and `BTreeMap`:
```rust
self.polys_map
    .into_par_iter()
    .map(|(label, p)| (label, B::self_to_coeff(domain, p)))
    .collect::<BTreeMap<_, _>>()
```

proofs/src/plonk/argument/verifier.rs
irakoton: I agree, if possible. It would make the code more stable when we edit it.

proofs/src/plonk/prover.rs
```diff
@@ -123,39 +123,41 @@ where
...
-            .into_iter()
-            .map(|(computed, commitment)| {
-                transcript.write(&commitment)?;
-                Ok(computed)
-            })
             .collect::<Result<Vec<_>, Error>>()?
     };
 
+    let logup_multiplicities_len = logup_multiplicities.len();
+    // TODO: Check this clone - see if we can remove it.
```
irakoton, line 150: I experimented on a branch: you can remove the clone, but it's a bit painful. The clone is quite big though, so that may be worth it, I don't know. Attempt [here](https://github.com/midnightntwrk/midnight-zk/commit/7a8d99d5aad9a2f7ced4982eb5ee2b6855fb37f7), explanation below.

You likely know it already, but if you just remove the clone (and use `into_iter` instead of `iter`) you get a borrow error for the second usage of `logup_multiplicities` later in the file. But the field used here (`multiplicities`) has basically a different status from the other fields of `logup_multiplicities`: you need it several times VS the other fields that are consumed by `compute_logderivative` and then dropped.

So one solution is to remove the `multiplicities` field from the struct, and manage it separately (have a function that computes it, and return a `(ComputedMultiplicities<F>, Polynomial<F, LagrangeCoeff>)` each time you would have returned a `ComputedMultiplicities<F>`. The change is not that big, but may be seen as a bit ugly I guess. But you do get rid of the clone this way.

proofs/src/poly/mod.rs
```diff
@@ -273,6 +307,13 @@ impl PolynomialRepresentation for LagrangeDoubleDeltaCoeff {
         )
     }
 
+    fn self_to_coeff<F: WithSmallOrderMulGroup<3>>(
+        _evaluation_domain: &EvaluationDomain<F>,
+        _poly: Polynomial<F, Self>,
+    ) -> Polynomial<F, Coeff> {
+        unimplemented!("Cannot convert LagrangeCoubleDelta basis to coefficient form.")
```
irakoton: LagrangeCoubleDelta -> LagrangeDoubleDelta

proofs/src/poly/kzg/commitment.rs
```diff
@@ -296,33 +296,49 @@ where
     }
 }
 
+/// Width of the little-endian prefix that frames a [`KZGMultiCommitment`] in a
+/// proof transcript, holding the number of polynomials the commitment covers.
+/// Only the points that follow it are hashed; the prefix delimits them for
+/// `Hashable::read`.
+pub(crate) const NB_POLYS_PREFIX_BYTES: usize = 4;
```
irakoton, line 294: Not a big fan of this, see comment below.

proofs/src/poly/kzg/commitment.rs
```diff
@@ -296,33 +296,49 @@ where
...
+        // matching `ProcessedSerdeObject::write`.
+        self.0.iter().flat_map(|c| c.to_input()).collect()
     }
 
     fn to_bytes(&self) -> Vec<u8> {
-        let mut bytes = Vec::new();
+        // The proof declares how many polynomials the group holds, so that
+        // `read` knows where the group ends. The count is not part of the
+        // hashed transcript; `read_commitment` checks it against the labels
+        // the verifying key expects.
```
irakoton, line 310: I found this comment suspicious, and after investigating it, I think there is a design issue with the way the `Hashable` is used.

What sounds wrong in the above comment is that the prover is writing some `u32` in the proof, but the verifier doesn't need it anyway since it gets it from the VK. I see that you need the value L329-330, but we're basically using the transcript for the prover to talk to itself. Honestly, having a full `u32` before each commitments for something that isn't even needed for verification seems a bit wasteful. 

If we decide to keep it, couldn't we at least use `u8`? I don't think there is any way to go above 255 commitments in a single phase. But I think the best design would be to pass this value through the code, instead of through the transcript.

irakoton: I made an attempt for this [here](https://github.com/midnightntwrk/midnight-zk/commit/8a567931638b732811e7be543f0fada7b94eeb90). Intuitively, I add a function `read_sized` to `Hashable`, unimplemented by default, and that is implemented for `KZGMultiCommitment` by performing `read` several times and wrapping up the result. This removes the need for the constant `NB_POLYS_PREFIX_BYTES`, the serialisation of the polynomial count, and the spurious checks the verifier performs.

irakoton (changes requested):
Thanks for the changes, this solves a lot of issues already. There are however a couple of points where I don't completely agree with the design. My main comments are a disagreement about the serialisation of the `u32`, and an answer to one of your TODOs (how to remove a nasty `clone`).

The comments and their impact are relatively local, but are still quite painful to describe, so I created a [branch](https://github.com/midnightntwrk/midnight-zk/tree/irakoton/pr515-suggested-changes) on top of yours, where the last two commits are each a suggested change for comments I made. The links to the commits in question are given in the associated comments.

proofs/src/poly/kzg/mod.rs
miguel-ambrona: Thanks so much 🫶

proofs/src/plonk/prover.rs
miguel-ambrona: The plan was to leave this clone for now. Any attempt to remove it is a distraction ATM.

proofs/src/poly/kzg/commitment.rs
miguel-ambrona: I agree, this should not stay. But for this refactor I prefer to do this and focus on getting everything else right.

irakoton (approved):

iquerejeta (approved):

iquerejeta (approved):

proofs/src/plonk/prover.rs
```diff
@@ -123,39 +123,41 @@ where
...
+    let phase1_polys_map = BTreeMap::from_iter(logup_multiplicities.iter().map(|c| {
+        (
+            PolynomialLabel::LogupMultiplicities(c.argument_index),
+            c.multiplicities.clone(),
+        )
+    }));
+
+    let phase1_committed =
+        argument::prover::Committed::commit::<CS, T>(params, phase1_polys_map, transcript)?;
+
```
gilescope: Claude says this drops #379's Δ/Δ² commit bases for multiplicities/aggregator leading to a +12% prove time when I ran it. Commit phase1 in Δ and phase2 in Δ², keep Lagrange polys for openings would make it 20% faster (at k=15) was it's suggestion to turn the slowdown into a win.

PR #525, opened 2026-09-16
MMR Gadget for Calypso

davidnevadoc:
- Exposes MMR Gagdet in Zkstdlib
- Adds conditional version of prefix check and membership check. These instructions return an assigned bit instead of asserting the condition. This causes certain regression when asserting. Around ~20% more rows.


circuits/src/mmr/mmr_gadget.rs
```diff
@@ -282,14 +278,19 @@ where
...
             // The sizes agree above height i and both MMRs have a mountain
             // here: their peaks must match directly.
             let direct_match = ng.and(
                 layouter,
                 &[agree[i + 1].clone(), a_bits[i].clone(), b_bits[i].clone()],
             )?;
-            ng.cond_assert_equal(layouter, &direct_match, &small.peaks[i], &big.peaks[i])?;
+            let peaks_equal = ng.is_equal(layouter, &small.peaks[i], &big.peaks[i])?;
+            let direct_ok = self.implies(layouter, &direct_match, &peaks_equal)?;
+            ok = ng.and(layouter, &[ok, direct_ok])?;
```
irakoton: IIUC, you have a kind of boolean accumulator `ok` that checks that peak equality is always verified. I would proceed the other way around, i.e., create a vector `failure` of `AssignedBit` that contain all the failures. The test "ok == true" thus becomes "`failure` is empty" (i.e., sum of all bits in `failure` validates `is_zero`).

The motivation is that it saves one row per iteration of the loop, i.e., `SIZE`. You would do something like
```rust
let peaks_differ = ng.not(layouter, &peaks_equal)?;
failures.push(ng.and(layouter, &[direct_match, peaks_differ])?);
```
which contains only 2 logical operators instead of 3 (counting `implies` as 2).

irakoton (commented):
I'll be commenting on some improvements regarding the implementation of the circuit. I'll probably propose a small PR on top of your stack that refactors the implementation to have small helpers, so that we can get back your 20% of regression while keeping the code clean.

circuits/src/mmr/mmr_gadget.rs
```diff
@@ -340,24 +376,30 @@ where
...
     /// # Unsatisfiable Circuit
     ///
-    /// If `elem` is not a member of `mmr`, or if the proof is malformed.
-    pub fn assert_membership<const SIZE: usize>(
+    /// If `leaf_index` does not fit in `SIZE` bits. The index is a hint the
+    /// prover is free to choose, so it is range-checked outright rather than
+    /// folded into the verdict; every semantic failure — wrong element, wrong
+    /// siblings, a height pointing at an absent mountain — returns `0`.
+    pub fn is_member<const SIZE: usize>(
         &self,
```
irakoton: IIUC, there is a prover supplied hint such that, when incorrect, makes the function return `False`. I.e., the function returning the assigned bit `True` is equivalent to `assert_membership`, but `False` may also mean "the witness is not well-formed" or something along these lines.

I don't know yet how this is weaved into Calypso, but what is the general intuition as to why this is fine? Is it something like these proofs being used in Calypso in a setting where you know "return `False`" is equivalent to "is not a member"? Or does the distinction simply does not matter?

Same remark and questions for the prefix variant.

circuits/src/mmr/mmr_gadget.rs
irakoton: Another comment is that, technically, this function does not agree with the CPU version anymore. In-circuit, you make the circuit unsatisfiable if `proof.leaf_index` is out of bounds, whereas off-circuit you return false. I think this should be patched, with either of two options:
1. Make the off-circuit version panic accordingly or return the adequate `Option<bool>`
2. Change the struct `MembershipProof` by making `proof.leaf_index` a `[F; SIZE]` instead of a `u64`, and the analogue modification in the assigned version. This way you are always in-bounds by construction.

I personally have a strong preference for the second version.

circuits/src/mmr/mmr_gadget.rs
```diff
@@ -55,6 +55,17 @@ pub struct AssignedMmr<F: CircuitField, const SIZE: usize> {
...
+    /// MMR state in its public input representation: size
+    /// followed by the peaks as SIZE+1 field elements. Agrees with
+    /// [Instantiable::as_public_input].
+    pub fn as_public_input(&self) -> Vec<AssignedNative<F>> {
+        let mut cells = vec![self.size.clone()];
+        cells.extend(self.peaks.iter().cloned());
+        cells
+    }
+}
+
```
irakoton: Why are we doing this? It seems you're implementing `as_public_input` and `constrain_as_public_input` as associated functions. Wouldn't it be better to simply implement `PublicInstructions`, to avoid the shadowing? I think that's why you had to rely on `as Instantiable` everywhere in the tests.

circuits/src/mmr/mmr_gadget.rs
irakoton: forgot to update the comment to the new function name

irakoton: forgot to update to the new function name

zk_stdlib/src/lib.rs
```diff
@@ -740,6 +746,29 @@ impl ZkStdLib {
...
+    /// (0..7u64).for_each(|i| mmr.append(F::from(i)));
+    ///
+    /// let mmr_gadget = chip.mmr();
+    /// let state = mmr_gadget.assign(layouter, Value::known(mmr.state()))?;
+    /// let elem: AssignedNative<F> = chip.assign(layouter, Value::known(F::from(3)))?;
+    /// let proof =
+    ///     mmr_gadget.assign_membership_proof(layouter, Value::known(mmr.prove_membership(3)))?;
+    /// mmr_gadget.assert_membership(layouter, &state, &elem, &proof)?;
+    /// # });
+    /// ```
```
irakoton: Nitpick, but this doc only talks about the previous interface.

irakoton: Maybe add this to the test
```rust
let ok = mmr_gadget.is_member(layouter, &state, &elem, &proof)?;
chip.assert_equal_to_fixed(layouter, &ok, true)?;
```

irakoton (dismissed):
Ok to the API change itself. I think there are better ways to handle the structure, see comments (and the comments on the previous PR). 

Note: All comments have been handled in #531.

PR #527, opened 2026-09-21
perf(circuits): derive Poseidon round constants once per field, not per hash

gilescope:
`SpongeCPU::init` calls `PreComputedRoundCPU::init()` on every hash, and the result depends only on the field. Cache it.

`bench_poseidon_cpu` hoists that setup out of its loop, so it never saw the cost. The new `poseidon_hash/hash_2` case benches the public entry point:

```
hash_2   35.6 µs → 13.2 µs   (permutation alone: 11.8 µs)
```

Adds `F: 'static` to `init()`; both implementors satisfy it.


DarrenOliveiro-Priestnall (approved):
I don't see any issues to this change from a security perspective.

